PR #2, opened 2019-12-10
Piece IO

ergastic:
### What
* Implementation of PieceIO and adjacent/dependent interfaces.
* Dependency on `filecoin-ffi`
* Makefile

### Pending
* ~Better unit test coverage~


pieceio/pieceio.go
```diff
@@ -0,0 +1,84 @@
...
+	commitment, err := pio.sectorCalculator.GeneratePieceCommitment(f, paddedSize)
+	if err != nil {
+		os.Remove(f.Name())
+		return nil, "", err
+	}
+	return commitment, filestore.Path(f.Name()), nil
+}
+
+func (pio *pieceIO) WritePayload(bs ReadStore, payloadCid cid.Cid, selector ipld.Node, w io.Writer) ([]byte, error) {
+	return nil, pio.carIO.WriteCar(context.Background(), bs, payloadCid, selector, w)
```
hannahhoward: I'm not sure value of this function at this point -- maybe delete it?

pieceio/pieceio.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+type pieceIO struct {
+	padReader        PadReader
+	carIO            CarIO
+	sectorCalculator SectorCalculator
+	tempDir          filestore.Path
+}
+
+func NewPieceIO(padReader PadReader, carIO CarIO, sectorCalculator SectorCalculator, tempDir filestore.Path) PieceIO {
```
hannahhoward: I wonder -- should PieceIO just have access to a Filestore directly and be called a PieceStore? And then we don't have to worry about getting things in and out of the filestore?

ergastic: In that case I'd like add a function to create a temporary file through the filestore.

PR #12, opened 2019-12-17
PieceIO improvements

ergastic:
## Changes in this PR:
* Improved test coverage for PieceIO.
* Changed PieceIO to use FileStore.
* Added CreateTemp method to FileStore.
* Enabled FileStore tests.


hannahhoward:
George can you rebase to resolve test issues?

filestore/filestore.go
```diff
@@ -62,5 +64,17 @@ func (fs fileStore) Store(p Path, src File) error {
 }
 
 func (fs fileStore) Delete(p Path) error {
-	return os.Remove(fs.filename(p))
+	err := os.Remove(string(p))
```
ingar: Seems like this could be changed to always delete something after constructing a path to within the filestore.  Safer also,  as there could be a relative path collision in and out of the filestore?

filestore/file.go
```diff
@@ -1,28 +1,35 @@
...
 }
 
 func (f fd) Path() Path {
-	return Path(f.filename)
+	return Path(f.Name())
 }
 
 func (f fd) Size() int64 {
-	info, err := os.Stat(f.filename)
+	info, err := os.Stat(fmt.Sprintf("%s%c%s", f.basepath, filepath.Separator, f.filename))
```
ingar: Seems like we are creating a full path a few times a few different ways; we could do...

```suggestion
	info, err := os.Stat(f.absPath())
```
if we had a function:

```go
func (f fd) absPath() string {
	return path.Join(f.basepath, f.filename)
}
```

filestore/filestore.go
```diff
@@ -48,19 +44,36 @@ func (fs fileStore) Create(p Path) (File, error) {
...
+	return Path(fs.filename(p)), nil
 }
 
 func (fs fileStore) Delete(p Path) error {
-	return os.Remove(fs.filename(p))
+	for idx := range fs.base {
+		if p[idx] != fs.base[idx] {
+			return fmt.Errorf("invalid base path for '%s' (expecting '%s')", string(p), fs.base)
+		}
+	}
```
ingar: There is a [`strings.HasPrefix`](https://golang.org/pkg/strings/#HasPrefix) that can do this.

PR #17, opened 2019-12-17
Network Abstraction For Retrieval Market

hannahhoward:
To Simplify RetrievalCode and improve testability, this creates a retrieval market network abstraction to handle the protocols

Interfaces are here, implementation to come


shared/statestore/store.go
```diff
@@ -0,0 +1,164 @@
+package statestore
+
+import (
+	"bytes"
```
hannahhoward: I think this may indicate a need to rebase again it's not part of go-fil-components any more.

shared/statestore/store_test.go
```diff
@@ -0,0 +1,38 @@
+package statestore
+
```
hannahhoward: also should not be in PR

retrievalmarket/impl/impl_types/types_cbor_gen.go
```diff
@@ -1,4 +1,4 @@
-package retrievalimpl
+package impl_types
```
hannahhoward: this is going away so there's no need to move it

filestore/filestore_test.go
```diff
@@ -3,6 +3,7 @@ package filestore
 import (
```
hannahhoward: I think this is another rebase that's in a weird place

shannonwells: This file is still in master.  The changes are because the linter complained about not error checking

shared_testutil/test_types.go
```diff
@@ -0,0 +1,109 @@
+package shared_testutil
```
hannahhoward: Neat!

retrievalmarket/network/libp2p_impl_test.go
```diff
@@ -0,0 +1,516 @@
...
+	case <-ctx.Done():
+		t.Error("response not received")
+	case <-done:
+	}
+
+	assert.Equal(t, qr, resp)
+}
+
+func TestQueryStreamSendReceiveOutOfOrderFails(t *testing.T) {
+	// send query, read response in handler - fails
```
hannahhoward: I had in mind just doing:
from: write query
to: read query response <- errors

Not the second step as well. It's fine that the test is actually testing more, but if it means you can get rid of the go routine I'd probably just do that.

retrievalmarket/network/libp2p_impl_test.go
```diff
@@ -0,0 +1,516 @@
...
+	select {
+	case <-ctx.Done():
+		t.Errorf("failed to receive messages")
+	case receivedPayment = <-dpyChan:
+	}
+
+	assert.Equal(t, dpy, receivedPayment)
+}
+
+func TestQueryStreamSendReceiveMultipleOutOfOrderFails(t *testing.T) {
```
hannahhoward: This is my own fault, but, honestly, this test doesn't make sense in retrospect.

I think the reason you're deadlocking without a go routine is the second WriteDealProposal never finishes in side the go routine. But cause there's no channel write at the end of the go routine, you just never see that.

It would be interesting to put a channel write at the end of the go routine and then try to read from it at the very end. I think you might find it never gets written to.

Assuming that's the case, I think this test gives a false impression of testing something it doesn't. Namely that all the writes after the first do not in fact pass. 

I'm not sure if there's a good way to resolve this, and I wonder if it's just unrealstic to trying to have this test -- in practice,
the handlers we write are likely to return as soon as they encounter an error, and there will not be multiple error reads in a row. 

So my honest, though non-blocking, recommendation is to delete this test and simplify the out-of-order query read test above to just test a single write/read, so that we don't have to deal with this. Sorry for having unneccesary work.

retrievalmarket/network/deal_stream.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	cborutil "github.com/filecoin-project/go-cbor-util"
+	"github.com/libp2p/go-libp2p-core/mux"
+	"github.com/libp2p/go-libp2p-core/peer"
+
+	"github.com/filecoin-project/go-fil-components/retrievalmarket"
+)
+
+type DealStream struct {
```
hannahhoward: NB: I wonder if this type should be private to the module. is there any downside there?

PR #20, opened 2019-12-19
filestore: use filepath pkg, track close err, lints

jsign:
Some minors while reading this pkg, most relevant:
- ~~use `filepath` to rely on stdlib~~
- since `Store` codeflow is simple enough, we can avoid defer to go the extra mile to see if closing fails.


mishmosh:
Thank you for this! We will review in the new year.

jsign:
Done.
Seems that in the meantime some other commit leveraged the `filepath` package, so that isn't a _new_ change anymore. (So updated the commit msg).

PR #25, opened 2020-01-03
Upgrade Query Protocol to Spec V0

hannahhoward:
# Goals

- Use the network abstractions being worked on by @shannonwells in the retrieval implementation
- Upgrade the Query Protocol to follow spec V0

# Implementation

- Make retrieval client/provider dependent on a network protocol implementation rather than directly on a libp2p host
- Add a stubbable mock implementation of the retrieval network (based on @shannonwells work)
- Update query handlers on the provider and client to use query protocol v0 from the spec
- Comment out deal code for the time being which is not v0
- Minor adds / changes to types
- Add tests for retrieval client/provider for queries

### DIFF
--- a/retrievalmarket/impl/client.go
+++ b/retrievalmarket/impl/client.go
@@ -94,16 +94,6 @@ func (c *client) Retrieve(ctx context.Context, pieceCID []byte, params retrieval
 
 	go func() {
 		evt := retrievalmarket.ClientEventError
-		//converted, err := cid.Cast(pieceCID)
-
-		/*if err == nil {
-			err = c.retrieveUnixfs(ctx, converted, tokenamount.Div(totalFunds, params.PricePerByte).Uint64(), totalFunds, miner, clientWallet, minerWallet)
-			if err == nil {
-				evt = retrievalmarket.ClientEventComplete
-				dealState.Status = retrievalmarket.DealStatusCompleted
-			}
-		}*/
-
 		c.notifySubscribers(evt, dealState)
 	}()
 
--- a/retrievalmarket/impl/client.go
+++ b/retrievalmarket/impl/client.go
@@ -27,9 +27,9 @@ import (
 var log = logging.Logger("retrieval")
 
 type client struct {
-	h    host.Host
-	bs   blockstore.Blockstore
-	node retrievalmarket.RetrievalClientNode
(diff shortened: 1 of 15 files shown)


retrievalmarket/impl/client.go
```diff
@@ -110,15 +84,15 @@ func (c *client) Retrieve(ctx context.Context, pieceCID []byte, params retrieval
 
 	go func() {
 		evt := retrievalmarket.ClientEventError
-		converted, err := cid.Cast(pieceCID)
+		//converted, err := cid.Cast(pieceCID)
```
shannonwells: Is there a reason to keep this commented-out code or did you mean to delete it?

retrievalmarket/impl/client.go
```diff
@@ -188,7 +163,7 @@ type clientStream struct {
 	verifier   BlockVerifier
 	bs         blockstore.Blockstore
 }
-
+*/
 /* This is the old retrieval code that is NOT spec compliant */
```
shannonwells: same re: commented out code

retrievalmarket/impl/client.go
hannahhoward: Please delete

retrievalmarket/impl/client.go
hannahhoward: do not delete in this case -- it's the whole old implementation -- need it for reference

retrievalmarket/impl/provider.go
```diff
@@ -159,26 +167,26 @@ type handlerDeal struct {
 }
 
 // TODO: Update for https://github.com/filecoin-project/go-retrieval-market-project/issues/7
-func (p *provider) handleDealStream(stream network.Stream) {
+func (p *provider) HandleDealStream(stream rmnet.RetrievalDealStream) {
 	defer stream.Close()
+	/*
```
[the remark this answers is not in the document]
hannahhoward: no please don't for now. I'm keeping it around as a reference.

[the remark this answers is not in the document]
shannonwells: ok I left that code alone, commenting out handlerDeal struct to please the linter, and rebased w/ master, plus some tidying and that's it.

PR #26, opened 2020-01-07
Use CAR file in generation of CommP

ingar:
## Summary
Use PieceIO and CAR file to generate CommP.

Resolves https://github.com/filecoin-project/go-storage-market/issues/13

### DIFF
--- a/storagemarket/impl/client.go
+++ b/storagemarket/impl/client.go
@@ -4,10 +4,16 @@ import (
 	"context"
 
 	"github.com/filecoin-project/go-data-transfer"
+	blockstore "github.com/ipfs/go-ipfs-blockstore"
+
+	"github.com/filecoin-project/go-fil-components/filestore"
+	"github.com/filecoin-project/go-fil-components/pieceio"
+	"github.com/filecoin-project/go-fil-components/pieceio/cario"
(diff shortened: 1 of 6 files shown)


storagemarket/impl/client.go
```diff
@@ -65,11 +75,22 @@ type clientDealUpdate struct {
 	mut      func(*ClientDeal)
 }
 
-func NewClient(h host.Host, dag ipld.DAGService, dataTransfer datatransfer.Manager, discovery *discovery.Local, deals *statestore.StateStore, scn storagemarket.StorageClientNode) *Client {
+func NewClient(h host.Host, bs blockstore.Blockstore, dataTransfer datatransfer.Manager, discovery *discovery.Local, deals *statestore.StateStore, scn storagemarket.StorageClientNode) (*Client, error) {
+	pr := padreader.NewPadReader()
+	carIO := cario.NewCarIO()
+	sectorCalculator := sectorcalculator.NewSectorCalculator("")
```
hannahhoward: non-blocking:
I think when we move to integration we may want implementations to setup PieceIO themselves and pass it as a depedency, along with the filestore.

I continue to think we should have pieceIO have a blockstore as well so we don't need it as a depedency here.

PR #28, opened 2020-01-09
Flaky test fix

shannonwells:
delete the tests


hannahhoward:
so we can make a decision to delete these tests, but as written, I'd assert these aren't testing anything any more. Once the go routine is there, the test just finishes before the statements inside the go routine are done. Perhaps there is just no way to make it work and we should delete these tests. clearly, out of order reads do not work :)

PR #36, opened 2020-01-10
RetrievalMarket: Query + Deal integration test, + bug fixes uncovered during writing the test

shannonwells:
### first test
- [x] construct two libp2p nodes
- [x] construct a retrieval network interface on top of each node
- [x] with one host:
         - make a mock RetrievalClientNode
         - make a  RetrievalClient
- [x] with the other:
        - make a mock RetrievalProviderNode
        - make a RetrievalProvider
- [x] make a query from the client to the provider and see if you get a response back

### second test

- [x] construct two libp2p nodes
- [x] construct a retrieval network interface on top of each node
- [x] with one host:
        * make a mock RetrievalClientNode
        * will want to record calls to create payment channel, allocate lane, setup voucher
        * make a  RetrievalClient
- [x] with the other:
        - Inject a unixFS file on the provider side to its blockstore (we did this in data transfer — it may actually be part of shared_testutil — if not check there)
        - make a mock RetrievalProviderNode - recommend the sealed blockstore just wrap the regular blockstore (they have the same interface) , also recommend processing payment vouchers by recording
        - make a RetrievalProvider
- [x] make a deal from the client to the provider and see if it completes
- [x] attempt to read out unixFS file from clients blockstore (see again datatransfer or shared_testutil)
- [ ] verify vouchers setup on client match those for pricePerByte, payment interval, & interval increase
- [x] verify vouchers setup on client are the same as those redeemed on provider
- [ ] downside is you won’t be able to get it to pass yet but I should have the provider side soon.


retrievalmarket/impl/integration_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	"github.com/filecoin-project/go-fil-markets/retrievalmarket"
+	retrievalimpl "github.com/filecoin-project/go-fil-markets/retrievalmarket/impl"
+	rmnet "github.com/filecoin-project/go-fil-markets/retrievalmarket/network"
+	"github.com/filecoin-project/go-fil-markets/shared/tokenamount"
+	"github.com/filecoin-project/go-fil-markets/shared/types"
+	tut "github.com/filecoin-project/go-fil-markets/shared_testutil"
+)
+
+func TestClientsCanTalkToEachOther(t *testing.T) {
```
hannahhoward: I would call this TestClientCanMakeQueryToProvider
NB

PR #43, opened 2020-01-14
Feat/find providers

shannonwells:
Closes [Zenhub Retrieval Market Project Issue #12](https://app.zenhub.com/workspaces/retrieval-market-5de558a85ea8f94a5e0882c1/issues/filecoin-project/go-retrieval-market-project/12)

Implement client.FindProviders method which just calls PeerResolver.GetPeers with the desired PieceCID; unit test with test PeerResolver

### DIFF
--- a/retrievalmarket/impl/client_test.go
+++ b/retrievalmarket/impl/client_test.go
@@ -2,11 +2,10 @@ package retrievalimpl_test
 
 import (
 	"context"
+	"errors"
 	"testing"
 
 	"github.com/filecoin-project/go-address"
-	"github.com/filecoin-project/go-data-transfer/testutil"
-	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-datastore"
 	dss "github.com/ipfs/go-datastore/sync"
 	bstore "github.com/ipfs/go-ipfs-blockstore"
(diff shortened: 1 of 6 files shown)


shannonwells:
> I have GetPeers taking a CID instead of bytes, because it is a commP...Would you have time to change that as part of this PR? If not it can be done seperately.

Meh, I think it's fine to change it before merging.

retrievalmarket/impl/client_test.go
```diff
@@ -105,10 +109,78 @@ func TestClient_Query(t *testing.T) {
...
+
+	t.Run("when there are no providers", func(t *testing.T) {
+		testResolver := testPeerResolver{peers: []retrievalmarket.RetrievalPeer{}}
+		c := retrievalimpl.NewClient(net, bs, &testRetrievalNode{}, &testResolver)
+		testCid := []byte("unimportant")
+		assert.Len(t, c.FindProviders(testCid), 0)
+	})
+}
+
+type testRetrievalNode struct {
```
hannahhoward: sorry hate to be a stickeler here, but unless there is a compelling reason, this should use the existing TestRetrievalClientNode -- there's no need to make a dupe here.

shannonwells: I had this in my branch before you merged yours to master, but agree there should not be two of them.  I overlooked that you had put one in.

shannonwells: I just missed that it had been moved out of this file when I rebased, is all.

PR #46, opened 2020-01-16
Support arbitrary dag retrieval

hannahhoward:
# Goals

This is a step on the way to supporting car files.
It replaces the UnixFS block reader and verifier with one that works on arbitrary dags using selectors.

# Implementation

- build a "traverser" which runs a selector traversal that stops at every block, allowing you to respond with the correct block to load in order to advance the query
- build a reader & verifier on top of that
- replace the unixfs reader and verifier


retrievalmarket/impl/blockio/verify.go
```diff
@@ -0,0 +1,66 @@
...
+type BlockVerifier interface {
+	Verify(context.Context, blocks.Block) (done bool, err error)
+}
+
+// OptimisticVerifier always verifies blocks
+type OptimisticVerifier struct {
+}
+
+// Verify always returns no error
+func (o *OptimisticVerifier) Verify(context.Context, blocks.Block) (bool, error) {
```
shannonwells: if this is actually a relic like you say let's delete it

retrievalmarket/impl/provider.go
```diff
@@ -198,31 +186,8 @@ func (p *provider) HandleDealStream(stream rmnet.RetrievalDealStream) {
...
-				return
-			}
-			size, err := fsr.Size()
-			if err != nil {
-				p.failDeal(&dealState, xerrors.Errorf("file %s didn't implement UnixfsReader", dealState.PayloadCID))
-				return
-			}
-			environment.size = uint64(size)
+		if environment.br == nil {
+			environment.br = blockio.NewSelectorBlockReader(cidlink.Link{Cid: dealState.PayloadCID}, loader)
```
shannonwells: 🎉

PR #48, opened 2020-01-17
Expose generic pieceio.GeneratePieceCommitment(Reader, size)

rvagg:
Testing the waters a bit here, let me know if this isn't worth pursuing. Happy to do more cleanup, tests, whatever else, I just need some guidance.

* Adds a static `func GeneratePieceCommitment(rd io.Reader, pieceSize uint64) ([]byte, uint64, error)` that's called by `PieceIO#GeneratePieceCommitment()`.
* Refactor use of PadReader
* Don't pad internal CAR files, wrap in PadReader and pass around paddedSize rather than using car.Size()

Ultimately it would allow us to externally generate CommP for an arbitrary `Reader` .. so, for example, we could run this in Lambda across pre-formed CAR files streamed from S3.

### DIFF
--- a/storagemarket/impl/provider_states.go
+++ b/storagemarket/impl/provider_states.go
@@ -107,7 +107,7 @@ func (p *Provider) verifydata(ctx context.Context, deal MinerDeal) (func(*MinerD
 	allSelector := ssb.ExploreRecursive(selector.RecursionLimitNone(),
 		ssb.ExploreAll(ssb.ExploreRecursiveEdge())).Node()
 
-	commp, file, err := p.pio.GeneratePieceCommitment(deal.Ref, allSelector)
+	commp, file, _, err := p.pio.GeneratePieceCommitment(deal.Ref, allSelector)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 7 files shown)


storagemarket/impl/provider_states.go
```diff
@@ -107,7 +107,7 @@ func (p *Provider) verifydata(ctx context.Context, deal MinerDeal) (func(*MinerD
 	allSelector := ssb.ExploreRecursive(selector.RecursionLimitNone(),
 		ssb.ExploreAll(ssb.ExploreRecursiveEdge())).Node()
 
-	commp, file, err := p.pio.GeneratePieceCommitment(deal.Ref, allSelector)
+	commp, file, _, err := p.pio.GeneratePieceCommitment(deal.Ref, allSelector)
```
hannahhoward: So this here is the main issue with this PR -- the file is read by someone else who expects it to have the length of the whole piece.

mikeal: Should we be storing .car files that are padded then?

In the offline dealflow, what we thought we should do is create .car files (no padding), calculate and store CommP, and then send those files to miners. Then the miners would need to calculate and generate CommP again to match. Would it be more efficient for us to just do the padding when we create the .car files, pass a padded file into a function like this, and then send the padded files to the miners?

hannahhoward: we resolved this else and this PR is merged :)

PR #49, opened 2020-01-17
Remove filestore as a go between with StorageMiner, pass direct io.reader to StorageMiner

hannahhoward:
# Goals

Many people have balked at the "Filestore" concept as a handoff between storage market and storage miner and requested we simply pass a direct io.reader from one to the other. This implements that change.

As I understand it, the filestore was originally introduced in specs on the theory that markets & miner might run on different machines, and that, rather than RPC over the whole piece, the filestore would act as a shared interface to a single storage staging area (a networked hard drive or perhaps other device). It seems to me while this concept might still have value, it's a concern best left to the code integrating storage market and storage miner, and the miner who wishes to customize that code.

# Implementation

While this PR does not delete the Filestore code, it moves all management of files internal to the module (also removing the requirement that the StorageMiner delete the files when it's done -- this is now managed internally as well).

I'm tagging @laser @acruikshank and @magik6k who will work with this in their own node implementations. Note @magik6k we are working on a replacement for sealed refs as well that works with CAR files that will be ready before we attempt to integrate this with Lotus.

The key line of concern to external stakeholders is this:

```diff
-	OnDealComplete(ctx context.Context, deal MinerDeal, piecePath string) (uint64, error)
+	OnDealComplete(ctx context.Context, deal MinerDeal, pieceSize uint64, pieceReader io.Reader) (uint64, error)
```

There is some other code to also make PieceIO more modular submitted by @rvagg so that external uses of go-fil-markets can generate commP from a CAR file they made themselves (there will be reasons to do this soon)

# For Discussion

One potential future inefficiency, in a scenario where one does wish to customize the handoff between markets and miner, is that we still have to write the piece file to disk internally, and if you were to write it to disk outside this module, it would be an inefficiency.

### DIFF
--- a/pieceio/pieceio.go
+++ b/pieceio/pieceio.go
@@ -30,10 +30,10 @@ func NewPieceIO(carIO CarIO, store filestore.FileStore, bs blockstore.Blockstore
 	return &pieceIO{carIO, store, bs}
 }
 
-func (pio *pieceIO) GeneratePieceCommitment(payloadCid cid.Cid, selector ipld.Node) ([]byte, filestore.File, uint64, error) {
+func (pio *pieceIO) GeneratePieceCommitment(payloadCid cid.Cid, selector ipld.Node) ([]byte, filestore.Path, uint64, error) {
 	f, err := pio.store.CreateTemp()
 	if err != nil {
-		return nil, nil, 0, err
+		return nil, "", 0, err
 	}
 	cleanup := func() {
 		f.Close()
@@ -42,20 +42,21 @@ func (pio *pieceIO) GeneratePieceCommitment(payloadCid cid.Cid, selector ipld.No
 	err = pio.carIO.WriteCar(context.Background(), pio.bs, payloadCid, selector, f)
 	if err != nil {
 		cleanup()
-		return nil, nil, 0, err
+		return nil, "", 0, err
 	}
 	pieceSize := uint64(f.Size())
 	_, err = f.Seek(0, io.SeekStart)
 	if err != nil {
 		cleanup()
-		return nil, nil, 0, err
+		return nil, "", 0, err
 	}
 	commitment, paddedSize, err := GeneratePieceCommitment(f, pieceSize)
 	if err != nil {
 		cleanup()
-		return nil, nil, 0, err
+		return nil, "", 0, err
 	}
-	return commitment, f, paddedSize, nil
+	_ = f.Close()
+	return commitment, f.Path(), paddedSize, nil
 }
 
 func GeneratePieceCommitment(rd io.Reader, pieceSize uint64) ([]byte, uint64, error) {
--- a/pieceio/pieceio.go
+++ b/pieceio/pieceio.go
@@ -2,7 +2,6 @@ package pieceio
 
 import (
 	"context"
-	"fmt"
 	"io"
 
 	"github.com/ipfs/go-cid"
@@ -10,13 +9,9 @@ import (
 	"github.com/ipld/go-ipld-prime"
 
 	"github.com/filecoin-project/go-fil-markets/filestore"
+	"github.com/filecoin-project/go-fil-markets/pieceio/padreader"
 	"github.com/filecoin-project/go-sectorbuilder"
 )
-
-type PadReader interface {
-	// PaddedSize returns the expected size of a piece after it's been padded
-	PaddedSize(size uint64) uint64
-}
 
 type CarIO interface {
 	// WriteCar writes a given payload to a CAR file and into the passed IO stream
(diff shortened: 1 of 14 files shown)


pieceio/pieceio.go
```diff
@@ -48,33 +42,29 @@ func (pio *pieceIO) GeneratePieceCommitment(payloadCid cid.Cid, selector ipld.No
...
 		cleanup()
-		return nil, nil, err
-	}
-	if uint64(padded) != remaining {
-		cleanup()
-		return nil, nil, fmt.Errorf("wrote %d byte of padding while expecting %d to be written", padded, remaining)
+		return nil, nil, 0, err
 	}
-	_, err = f.Seek(0, io.SeekStart)
+	commitment, paddedSize, err := GeneratePieceCommitment(f, pieceSize)
```
laser: I see that you've fseek'ed the file back to 0 after the call to `pio.carIO.WriteCar`. I think it would be good manners to do the same after `GeneratePieceCommitment`.

pieceio/pieceio.go
```diff
@@ -26,20 +21,19 @@ type CarIO interface {
...
 }
 
-func NewPieceIO(padReader PadReader, carIO CarIO, store filestore.FileStore, bs blockstore.Blockstore) PieceIO {
-	return &pieceIO{padReader, carIO, store, bs}
+func NewPieceIO(carIO CarIO, store filestore.FileStore, bs blockstore.Blockstore) PieceIO {
+	return &pieceIO{carIO, store, bs}
 }
 
-func (pio *pieceIO) GeneratePieceCommitment(payloadCid cid.Cid, selector ipld.Node) ([]byte, filestore.File, error) {
+func (pio *pieceIO) GeneratePieceCommitment(payloadCid cid.Cid, selector ipld.Node) ([]byte, filestore.File, uint64, error) {
```
laser: You're writing the CAR to a tempfile here. Some Linux distros (like [Arch](https://wiki.archlinux.org/index.php/Tmpfs)) use a ramdisk for `/tmp`. Does `CreateTemp()` write to `/tmp`? If so, I would expect this to be the source of OOMs for miners committing large sectors. 

A potential solution would be to modify `pieceIO#GeneratePieceCommitment` to return a `Reader` instead of a `File`. You could use (`os.Pipe`)[https://golang.org/pkg/os/#Pipe], passing the write side of the pipe to `pio.carIO.WriteCar` in a goroutine and return the read side of the pipe.

pieceio/pieceio.go
laser: (if you stick with the fseekable approach)

storagemarket/impl/provider_states.go
```diff
@@ -195,12 +202,24 @@ func (p *Provider) staged(ctx context.Context, deal MinerDeal) (func(*MinerDeal)
...
 func (p *Provider) sealing(ctx context.Context, deal MinerDeal) (func(*MinerDeal), error) {
 	// TODO: consider waiting for seal to happen
+	cb := func(err error) {
+		select {
+		case p.updated <- minerDealUpdate{
+			newState: storagemarket.DealComplete,
+			id:       deal.ProposalCid,
+			err:      err,
+		}:
+		case <-p.stop:
```
laser: I can't tell from looking at this PR where it is that you send a value to `p.stop` - but I do see that you've got at least one anonymous goroutine (elsewhere in the file) which uses receipt of a value on `p.stop` to signal exit. 

If it's the case that a single termination value on `p.stop` is sent - then I fear that you'll end up with orphaned goroutines (only one will receive the termination value).

pieceio/pieceio.go
hannahhoward: golang's temp file is not writing to a ram disk -- it puts it in the directory you specify

hannahhoward: I will look at piping.

storagemarket/impl/provider_states.go
hannahhoward: p.stop sends by being closed, which means it will read multiple times.

laser: Nice. I didn't know that about Golang channels.

storagemarket/impl/provider_states.go
```diff
@@ -179,7 +185,8 @@ func (p *Provider) staged(ctx context.Context, deal MinerDeal) (func(*MinerDeal)
 			Ref:         deal.Ref,
 			DealID:      deal.DealID,
 		},
-		string(deal.PiecePath),
+		paddedSize,
+		paddedReader,
```
laser: Very nice!

laser (approved):
Looks good to me, modulo the temp file thing. Thanks for making this change.

PR #50, opened 2020-01-21
Sync up DealState to match spec

ingar:
## Summary
Rename `DealState` to `StorageDealStatus`, add the states from the spec, and map them as closely as possible to the old `DealState` enum.

### Notes
There are some unused states, and the provider and client states are not broken up according to the states in the spec (leaving some unused states).  This will be addressed when we figure out what is going on with the state machines for the client and provider deals.

Resolves https://github.com/filecoin-project/go-storage-market/issues/21

### DIFF
--- a/storagemarket/types.go
+++ b/storagemarket/types.go
@@ -27,10 +27,10 @@ type Balance struct {
 	Available tokenamount.TokenAmount
 }
 
-type DealState = uint64
+type StorageDealStatus = uint64
 
 const (
-	DealUnknown  = DealState(iota)
+	DealUnknown  = StorageDealStatus(iota)
 	DealRejected // Provider didn't like the proposal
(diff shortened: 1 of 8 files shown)


storagemarket/types.go
```diff
@@ -27,43 +27,56 @@ type Balance struct {
...
+	StorageDealProposalSigned
+	StorageDealPublished
+	StorageDealCommitted
+	StorageDealActive
+	StorageDealFailing
+	StorageDealRecovering
+	StorageDealExpired
+	StorageDealNotFound
 
 	// Internal
```
hannahhoward: ^^ NB: These are all "internal" in the sense that they are not actually states that exist outside these markets. In any case, I don't see much difference.

PR #55, opened 2020-01-23
Switch interfaces for retrieval to use Payload CIDs. NO IMPLEMENTATION

hannahhoward:
# Goals

To support both Lotus and GFC integration, we are switching away from using pieceCIDs for retrieval.
This PR changes the names and types on the expected interfaces, thought importantly, it leaves the implementation the same under the hood (treating payloadCIDs as if they were pieceCIDs). The fix to actually implement PayloadCID based retrieval will come in the next PR.

### DIFF
--- a/retrievalmarket/network/libp2p_impl_test.go
+++ b/retrievalmarket/network/libp2p_impl_test.go
@@ -12,7 +12,6 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/filecoin-project/go-address"
-	"github.com/filecoin-project/go-data-transfer/testutil"
 	"github.com/filecoin-project/go-fil-markets/retrievalmarket"
 	"github.com/filecoin-project/go-fil-markets/retrievalmarket/network"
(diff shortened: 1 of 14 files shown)


retrievalmarket/network/libp2p_impl_test.go
```diff
@@ -113,10 +112,10 @@ func TestQueryStreamSendReceiveMultipleSuccessful(t *testing.T) {
 	qs, err := nw1.NewQueryStream(td.Host2.ID())
 	require.NoError(t, err)
 
-	testCid := testutil.GenerateCids(1)[0]
+	testCid := shared_testutil.GenerateCids(1)[0]
```
shannonwells: ❤️

shared_testutil/testutil.go
```diff
@@ -0,0 +1,83 @@
+package shared_testutil
```
shannonwells: thanks for this

hannahhoward:
@ingar I'll handle changes to the piece store

PR #57, opened 2020-01-24
Properly Implement Retrieval Lookups Based on CIDs

hannahhoward:
# Goals

Allow Retrieval deals to work with actual Payload CIDs

# Implementation

-- add a second mapping to the PieceStore that translates from a CID to pieces its a part of
-- support unsealling pieces from CIDs as needed (theoretically a dag could span multiple pieces in the future where pieces are not whole dags -- not super important though)
-- add very preliminary writes of CID -> Piece mapping on storage provider side
    --- IMPORTANT: Not included is mapping any CIDs in the IPLD graph other than the root to Piece, or recording where in the piece blocks actually live. This can definitely be done by recording this information as we write the CAR file. This is definitely temporary implementation meant to get us to working retrieval by payload CID fast, with allowing retrievals based off of CIDs deeper in the dag later.

### DIFF
--- a/piecestore/piecestore.go
+++ b/piecestore/piecestore.go
@@ -1,87 +1,122 @@
 package piecestore
 
 import (
+	"bytes"
 	"fmt"
 
 	"github.com/filecoin-project/go-statestore"
+	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-datastore"
 	"github.com/ipfs/go-datastore/namespace"
 )
 
-var DSPrefix = "/storagemarket/pieces"
+var DSPiecePrefix = "/storagemarket/pieces"
+var DSCIDPrefix = "/storagemarket/cid-infos"
 
 func NewPieceStore(ds datastore.Batching) PieceStore {
 	return &pieceStore{
-		store: statestore.New(namespace.Wrap(ds, datastore.NewKey(DSPrefix))),
+		pieces:   statestore.New(namespace.Wrap(ds, datastore.NewKey(DSPiecePrefix))),
+		cidInfos: statestore.New(namespace.Wrap(ds, datastore.NewKey(DSCIDPrefix))),
 	}
 }
 
 type pieceStore struct {
-	store *statestore.StateStore
+	pieces   *statestore.StateStore
+	cidInfos *statestore.StateStore
 }
 
 func (ps *pieceStore) AddDealForPiece(pieceCID []byte, dealInfo DealInfo) error {
 	// Do we need to de-dupe or anything here?
 	return ps.mutatePieceInfo(pieceCID, func(pi *PieceInfo) error {
+		for _, di := range pi.Deals {
+			if di == dealInfo {
+				return nil
+			}
+		}
 		pi.Deals = append(pi.Deals, dealInfo)
 		return nil
 	})
 }
 
-func (ps *pieceStore) AddBlockInfosToPiece(pieceCID []byte, blockInfos []BlockInfo) error {
-	// Do we need to de-dupe or anything here?
-	return ps.mutatePieceInfo(pieceCID, func(pi *PieceInfo) error {
-		pi.Blocks = blockInfos
-		return nil
-	})
-}
-
-func (ps *pieceStore) HasBlockInfo(pieceCID []byte) (bool, error) {
-	pi, err := ps.GetPieceInfo(pieceCID)
-	if err != nil {
-		return false, err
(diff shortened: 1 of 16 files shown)


storagemarket/impl/provider_states.go
```diff
@@ -221,15 +222,21 @@ func (p *Provider) sealing(ctx context.Context, deal MinerDeal) (func(*MinerDeal
...
-        }
+	err := p.fs.Delete(deal.PiecePath)
+	if err != nil {
+		return nil, err
+	}
 	sectorID, offset, length, err := p.spn.LocatePieceForDealWithinSector(ctx, deal.DealID)
 	if err != nil {
 		return nil, err
 	}
+	// TODO: Record actual block locations for all CIDs in piece by improving car writing
```
shannonwells: SHould this be captured in an issue?

storagemarket/impl/provider_states.go
```diff
@@ -221,15 +222,21 @@ func (p *Provider) sealing(ctx context.Context, deal MinerDeal) (func(*MinerDeal
...
+	err = p.pieceStore.AddPieceBlockLocations(deal.Proposal.PieceRef, map[cid.Cid]piecestore.BlockLocation{
+		deal.Ref: {},
```
ingar: What is needed to store the root CID location here?

ingar:
The idea is we can store an index to arbitrary blocks in the IPLD graph of a deal payload, because a CAR file is just blocks that are concatenated together?

piecestore/piecestore_test.go
```diff
@@ -9,51 +9,164 @@ import (
...
+		dealInfo := piecestore.DealInfo{
+			DealID:   rand.Uint64(),
+			SectorID: rand.Uint64(),
+			Offset:   rand.Uint64(),
+			Length:   rand.Uint64(),
+		}
+		err := ps.AddDealForPiece(pieceCid, dealInfo)
+		assert.NoError(t, err)
+
+		pi, err := ps.GetPieceInfo(pieceCid)
```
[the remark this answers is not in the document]
hannahhoward: I just follow past of least resistance to get no lint-stage errors

storagemarket/impl/provider_states.go
hannahhoward: yes, it will be soon, that's the intent.

piecestore/piecestore.go
```diff
@@ -1,87 +1,126 @@
...
 }
 
 type pieceStore struct {
-	store *statestore.StateStore
+	pieces   *statestore.StateStore
+	cidInfos *statestore.StateStore
 }
 
 func (ps *pieceStore) AddDealForPiece(pieceCID []byte, dealInfo DealInfo) error {
 	// Do we need to de-dupe or anything here?
```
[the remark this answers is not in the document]
hannahhoward: ok!

storagemarket/impl/provider_states.go
hannahhoward: even for the root, we need to figure out where it sits within the CAR file, so it will be addressed at the same time we address the todo above

hannahhoward:
@ingar -- basically yes -- when we write the car, we can basically record for any CID:
- the piece its in
   - the relative offset within the piece
   - the size of the block
And then when we combine that with pieces mapping, we can effectively get, for any CID:
- a sector it's in
- an offset within the sector
- a length of bytes within the sector.

And theoretically we might start partial unsealling even pieces based on that, except that requires unsealing to be super fast which... I'm not sure where things are at with that.

retrievalmarket/impl/blockunsealing/blockunsealing.go
```diff
@@ -92,3 +87,22 @@ func (lu *loaderWithUnsealing) attemptUnseal() error {
 
 	return nil
 }
+
+func (lu *loaderWithUnsealing) firstSuccessfulUnseal(cidInfo piecestore.CIDInfo) (io.ReadCloser, error) {
```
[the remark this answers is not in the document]
hannahhoward: so the logic changed a bit here. theoretically since we're basing our queries on payload CIDs, I guess it's possible that a query could span multiple pieces now (if a piece was a partial but not complete DAG -- something we intend to support I believe). 

The point of this function is: for a given Cid, find the first piece that it's contained in that we can successfully unseal.

since we unseal and read the whole piece, in theory this function should not get called unless the query we make somehow crosses a piece boundary, which again, I think is possible.

If unsealing fails completely the first time, it's probably alright cause the loader calling code will immediately error and stop the retrieval deal.

PR #105, opened 2020-01-31
Add Id to StorageDeal struct

ingar:
## Summary
It would be useful to have the deal id when listing deals from the chain.  This adds the Deal ID to the structure returned from `ListClientDeals` and `ListProviderDeals`.

### DIFF
--- a/storagemarket/types.go
+++ b/storagemarket/types.go
@@ -152,6 +152,7 @@ type StorageDeal struct {
 	StoragePricePerEpoch tokenamount.TokenAmount
 	StorageCollateral    tokenamount.TokenAmount
 	ActivationEpoch      uint64 // 0 = inactive
+	Id                   DealID
 }
 
 type StateKey interface {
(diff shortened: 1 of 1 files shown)


anorth:
I can't tell if StorageDeal is intended to be an on-chain type, but https://github.com/filecoin-project/go-fil-markets/issues/106. FYI @hannahhoward

ingar:
This is our own type that closely mirrors the on-chain StorageDeal; but if we can directly use the spec-actors version it could simplify things.

hannahhoward:
@ingar it is an on-chain type and that's how it's treated in the Lotus integration. 

I would prefer not to modify things that are on-chain types in implementations

hannahhoward:
I would prefer the strategy in #106 but we need to figure out timing. Let's not merge this now.

hannahhoward:
Closing in favor of spec-actors type import

PR #121, opened 2020-02-12
Update go-sectorbuilder to latest that uses specs-actors types

hannahhoward:
# Goals

We updated our internal types, but some of our dependencies updated as well -- this updates to use the go-sectorbuilder that uses specs-actors types

# Implementation

Update -- FFI, go-sectorbuilder, & go-padreader

Also use abi.UnpaddedPieceSize where appropriate, and make appropriate conversion in deal PieceSize calculation


hannahhoward:
Padded() takes the pieceSize w/o bit padding (but w/ power of two 2 padding at the end) and returns the size w/ padding

PR #136, opened 2020-03-05
Storage Client Statemachine Refactor

hannahhoward:
# Goals

Make the storage client
- better commented
- easier to navigate and look at code for
- more well tested

# Implementation

- Setup a client state machine using all the FSM module in go-statemachine
- Test all state handlers with unit tests
- Define events for everything that happens in the storage client module
- Add additional infrastructure to support more testing
- Setup networking mocks for streams


storagemarket/impl/client.go
```diff
@@ -88,93 +75,24 @@ func NewClient(
...
-		c.handle(ctx, deal, c.new, storagemarket.StorageDealProposalAccepted)
-	case storagemarket.StorageDealProposalAccepted:
-		c.handle(ctx, deal, c.accepted, storagemarket.StorageDealStaged)
-	case storagemarket.StorageDealStaged:
-		c.handle(ctx, deal, c.staged, storagemarket.StorageDealSealing)
-	case storagemarket.StorageDealSealing:
-		c.handle(ctx, deal, c.sealing, storagemarket.StorageDealNoUpdate)
-		// TODO: StorageDealActive -> watch for faults, expiration, etc.
-	}
+func (c *Client) Run(ctx context.Context) {
```
ingar: I guess this can boot up in-progress deals when restarting the node?

storagemarket/impl/clientstates/client_fsm.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/filecoin-project/go-fil-markets/storagemarket"
+	"github.com/filecoin-project/go-statemachine/fsm"
+	"github.com/filecoin-project/specs-actors/actors/abi"
+	"github.com/ipfs/go-cid"
+	"golang.org/x/xerrors"
+)
+
+// ClientEvents are the events that can happen in a storage client
+var ClientEvents = fsm.Events{
+	fsm.Event(storagemarket.ClientEventOpen).
```
ingar: I don't quite remember why we send an event like this.  Is it because we need an arbitrary event for the FSM to start running?  Doesn't `Begin()` do that?

storagemarket/impl/clientstates/client_fsm.go
```diff
@@ -0,0 +1,105 @@
...
+		From(storagemarket.StorageDealUnknown).To(nil),
```
ingar: Should this be `.ToNoChange()`?

storagemarket/impl/clientstates/client_fsm.go
```diff
@@ -0,0 +1,105 @@
...
+	fsm.Event(storagemarket.ClientEventFundsEnsured).
+		From(storagemarket.StorageDealUnknown).To(storagemarket.StorageDealFundsEnsured),
+	fsm.Event(storagemarket.ClientEventWriteProposalFailed).
+		From(storagemarket.StorageDealFundsEnsured).To(storagemarket.StorageDealError).
+		Action(func(deal *storagemarket.ClientDeal, err error) error {
+			deal.Message = xerrors.Errorf("sending proposal to storage provider failed: %w", err).Error()
+			return nil
+		}),
+	fsm.Event(storagemarket.ClientEventDealProposed).
+		From(storagemarket.StorageDealFundsEnsured).To(storagemarket.StorageDealValidating),
```
ingar: Ugh, I wonder what the race condition recovery looks like here with multiple simultaneous deals..

storagemarket/impl/clientstates/client_states.go
```diff
@@ -0,0 +1,122 @@
...
+
+	proposal := network.Proposal{DealProposal: &deal.ClientDealProposal, Piece: deal.DataRef}
+	if err := s.WriteDealProposal(proposal); err != nil {
+		return ctx.Trigger(storagemarket.ClientEventWriteProposalFailed, err)
+	}
+
+	return ctx.Trigger(storagemarket.ClientEventDealProposed)
+}
+
+// VerifyDealResponse reads and verifies the response from the provider to the proposed deal
```
ingar: Should read, "VerifyDealResponse waits, waits, and waits, and waits some more, and then reads and verifies..."  :)

storagemarket/impl/client.go
hannahhoward: yes, "in the future" :)

storagemarket/impl/clientstates/client_fsm.go
hannahhoward: no Begin() does not do that -- it sets an initial state, but there is no implied initial event

storagemarket/impl/clientstates/client_fsm.go
hannahhoward: right. #whenYouForGetYourOwnDSL

storagemarket/impl/clientstates/client_fsm.go
hannahhoward: well the upside is the provider validates it too... so maybe they'll catch it? But you are right that the logic ought to be ensure funds for all deals in progress... dunno what that looks like.

storagemarket/impl/clientstates/client_states.go
hannahhoward: omg... I am very ready to send that intent to accept a lot earlier and switch up the data transfer initiation... just gotta finish this here refactor first.

storagemarket/impl/client.go
```diff
@@ -223,30 +137,32 @@ func (c *Client) Start(ctx context.Context, p ClientDealProposal) (cid.Cid, erro
 		return cid.Undef, xerrors.Errorf("getting proposal node failed: %w", err)
 	}
 
-	s, err := c.net.NewDealStream(p.MinerID)
-	if err != nil {
-		return cid.Undef, xerrors.Errorf("connecting to storage provider failed: %w", err)
+	deal := &storagemarket.ClientDeal{
+		ProposalCid:        proposalNd.Cid(),
+		ClientDealProposal: *clientDealProposal,
+		State:              storagemarket.StorageDealUnknown,
```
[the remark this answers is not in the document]
hannahhoward: I agree will fix in future PR

PR #145, opened 2020-03-11
Refactor Storage Provider to FSM Module

hannahhoward:
# Goals

Convert the storage provider to the FSM module, add many unit tests, generally refactor to make code more manageable and legible

# Implementation

- Refactor all provider states to the FSM module
- Factor our some common utility functions
- Factor out storage ask code from storage provider -- this is a fairly discreet unity of functionality which makes the storage provider code much larger
- Factor out common code on client & provider used to manage connections
- Add unit tests to: new provider FSM code, utility functions, factored out modules


hannahhoward:
@ingar -- my next step is to finally do that conditional acceptance ticket (where the provider sends conditional accepts and the client creates the data transfer)

PR #149, opened 2020-03-13
add TipSet identity-producing method to various Node interfaces

laser:
This PR makes incremental progress towards [completing this ZenHub issue](https://app.zenhub.com/workspaces/markets-shared-components-5daa144a7046a60001c6e253/issues/filecoin-project/go-fil-markets/148).

## Why does this PR exist?

When a provider or client needs information from the chain, they need to provide a chain state identifier to which their query is scoped. Due to potential reorgs, chain height does not serve as identity for a tipset.

## What's in this PR?

This PR changes the signature and name of the existing chain state identity-acquiring method, and adds that method to all `*Node` interfaces which query for (or will query for, in the short future) chain state.

### DIFF
--- a/storagemarket/testnodes/testnodes.go
+++ b/storagemarket/testnodes/testnodes.go
@@ -19,21 +19,6 @@ import (
 
 // Below fake node implementations
 
-// TestStateKey is just a stubbed state key that returns a preset height
-type TestStateKey struct {
-	Epoch abi.ChainEpoch
-	Token shared.TipSetToken
-}
-
-func (k *TestStateKey) TipSetToken() shared.TipSetToken {
-	return k.Token
-}
-
-// Height returns the value specified by Epoch
-func (k *TestStateKey) Height() abi.ChainEpoch {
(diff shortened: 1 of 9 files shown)


storagemarket/testnodes/testnodes.go
```diff
@@ -96,12 +92,14 @@ type FakeCommonNode struct {
...
+// GetChainHead returns the state id in the storage market state
+func (n *FakeCommonNode) GetChainHead(ctx context.Context) (shared.TipSetToken, abi.ChainEpoch, error) {
 	if n.MostRecentStateIDError == nil {
-		return n.SMState.StateKey(), nil
+		key, epoch := n.SMState.StateKey()
+		return key, epoch, nil
 	}
-	return &TestStateKey{}, n.MostRecentStateIDError
+
+	return []byte{}, 0, n.MostRecentStateIDError
```
hannahhoward: we might as well rename this symbol to GetChainHeadError

Issue #148, opened 2020-03-13
RetrievalProviderNode, StorageProviderNode, RetrievalClientNode, and StorageClientNode should provide a TipSet identifier when querying blockchain


laser:
## Why does this ticket exist?

When a provider or client needs some information from the chain, they need to provide the a tipset identity such that a sequence of queries which are _intended_ to be against a single chain state are _guaranteed_ to be against the same chain state. 

Note: A tipset height does not serve as identity due to re-orgs (tipset X at height 10 may be reorged out of existence, replaced with tipset Y).

## Acceptance Criteria

- each `*Node` interface must expose a method used to get the current chain head's tipset identity
- each `Get*` method (used to query chain state) must accept a tipset identity to which the query is scoped

PR #153, opened 2020-03-17
send state identifiery when getting miner worker address

laser:
Makes incremental progress towards https://github.com/filecoin-project/go-fil-markets/issues/148.

### DIFF
--- a/retrievalmarket/impl/provider.go
+++ b/retrievalmarket/impl/provider.go
@@ -171,7 +171,15 @@ func (p *provider) HandleQueryStream(stream rmnet.RetrievalQueryStream) {
 		MaxPaymentIntervalIncrease: p.paymentIntervalIncrease,
 	}
 
-	paymentAddress, err := p.node.GetMinerWorker(context.TODO(), p.minerAddress)
+	ctx := context.TODO()
+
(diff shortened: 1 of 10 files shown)


laser:
> does an address lookup really change based on the TipSet key?

Yep, see [here](https://github.com/filecoin-project/specs-actors/blob/2a954f25482f3e543cd663e2f4710561a6e940b0/actors/builtin/miner/miner_state.go#L33). A storage miner actor's worker address may be X at block height H and X' at H'. Passing the tipset key allows disambiguation when chain queries from the provider span multiple chain heads.

> just curious -- why the method rename?

I changed the method name to match what's in go-storage-miner. I could change what's in go-storage-miner to match what's in go-fil-markets, I suppose - but I'm already making edits to this codebase. What do you prefer?

laser:
Oops. I linked to the wrong part of specs-actors. See [this link](https://github.com/filecoin-project/specs-actors/blob/3f5510bf61303d1f629f8fce26ebe3194c911f93/actors/builtin/miner/miner_actor.go#L117) for more info.

hannahhoward:
all good! LGTM

PR #164, opened 2020-03-17
retrieval client node interface passes tipset identifier to node

laser:
## Why does this PR exist?

This PR makes incremental progress towards the larger goal of providing a tipset identifier to the node to which it scopes all chain queries.

### DIFF
--- a/retrievalmarket/impl/clientstates/client_states.go
+++ b/retrievalmarket/impl/clientstates/client_states.go
@@ -20,7 +20,12 @@ type ClientDealEnvironment interface {
 
 // SetupPaymentChannel sets up a payment channel for a deal
 func SetupPaymentChannel(ctx fsm.Context, environment ClientDealEnvironment, deal rm.ClientDealState) error {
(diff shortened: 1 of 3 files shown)


retrievalmarket/impl/clientstates/client_states.go
```diff
@@ -20,7 +20,12 @@ type ClientDealEnvironment interface {
 
 // SetupPaymentChannel sets up a payment channel for a deal
 func SetupPaymentChannel(ctx fsm.Context, environment ClientDealEnvironment, deal rm.ClientDealState) error {
-	paych, err := environment.Node().GetOrCreatePaymentChannel(ctx.Context(), deal.ClientWallet, deal.MinerWallet, deal.TotalFunds)
+	tok, _, err := environment.Node().GetChainHead(ctx.Context())
+	if err != nil {
+		return ctx.Trigger(rm.ClientEventPaymentChannelErrored, err)
```
laser: @hannahhoward Should I create a new error for this state - or will the one which I've used suffice?

hannahhoward: I think it's fine

laser: okie dokie

PR #171, opened 2020-03-25
Feat/configure start epoch buffer

### LINKED ISSUE #139: Storage Provider Can Configure Minimum StartEpoch in future time

ingar:
## Summary
Add a configurable `DealAcceptanceBuffer`, which can be used by storage providers to adjust how far in the future a deal's `StartEpoch` must be in order to be accepted.  This is to account for time required for data transfer, deal publishing, sealing, committing, etc.

Resolves #139

### DIFF
--- a/storagemarket/impl/provider.go
+++ b/storagemarket/impl/provider.go
@@ -30,6 +30,7 @@ import (
 )
 
 var ProviderDsPrefix = "/deals/provider"
+var DefaultDealAcceptanceBuffer = abi.ChainEpoch(100)
 var _ storagemarket.StorageProvider = &Provider{}
 
 // Provider is a storage provider implementation
@@ -92,7 +93,7 @@ func NewProvider(net network.StorageMarketNetwork, ds datastore.Batching, bs blo
 		storedAsk:            storedAsk,
 		actor:                minerAddress,
 		dataTransfer:         dataTransfer,
-		dealAcceptanceBuffer: abi.ChainEpoch(100),
+		dealAcceptanceBuffer: DefaultDealAcceptanceBuffer,
 	}
 
 	deals, err := fsm.New(namespace.Wrap(ds, datastore.NewKey(ProviderDsPrefix)), fsm.Parameters{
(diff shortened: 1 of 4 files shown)


storagemarket/impl/provider.go
```diff
@@ -265,6 +273,20 @@ func (p *Provider) HandleAskStream(s network.StorageAskStream) {
...
+	for _, option := range options {
+		option(p)
+	}
+}
+
+func (p *Provider) DealAcceptanceBuffer() abi.ChainEpoch {
+	return p.dealAcceptanceBuffer
+}
+
+func (p *Provider) UniversalRetrievalEnabled() bool {
```
hannahhoward: What is the value of these functions if they are not added to the StorageProvider interface type? Should they be? I'm not sure they should?

Oh, nm, you're using them for testing. Makes sense. +1

storagemarket/impl/provider.go
```diff
@@ -339,4 +361,8 @@ func (p *providerDealEnvironment) Disconnect(proposalCid cid.Cid) error {
 	return p.p.conns.Disconnect(proposalCid)
 }
 
+func (p *providerDealEnvironment) DealAcceptanceBuffer() abi.ChainEpoch {
```
hannahhoward: So I've gone back and forth a bunch on this, but I wonder if we should throw out the providerDealEnvironment and just make these methods on the Provider, cause we have the same aforementioned feature above -- the methods aren't accessible assuming you just have the storage provider interface. Non-blocking, but what do you think?

storagemarket/impl/provider.go
```diff
@@ -73,16 +82,17 @@ func NewProvider(net network.StorageMarketNetwork, ds datastore.Batching, bs blo
...
+		proofType:            rt,
+		spn:                  spn,
+		fs:                   fs,
+		pio:                  pio,
+		pieceStore:           pieceStore,
+		conns:                connmanager.NewConnManager(),
+		storedAsk:            storedAsk,
+		actor:                minerAddress,
+		dataTransfer:         dataTransfer,
+		dealAcceptanceBuffer: abi.ChainEpoch(100),
```
hannahhoward: NIT: can we put this in a constant at the top of the file? I prefer not to have hardcoded numbers in the constructor

storagemarket/impl/provider.go
ingar: I like implementing the interface on our provider impl; it would remove a bunch of indirection.

PR #172, opened 2020-03-25
send tipset identifier to node when interacting with chain

laser:
## Why does this PR exist?

This is the last (hopefully) PR in a chain of PRs which pass chain state identifiers to methods on the `Node*` interfaces.

### DIFF
--- a/storagemarket/testnodes/testnodes.go
+++ b/storagemarket/testnodes/testnodes.go
@@ -109,7 +109,7 @@ func (n *FakeCommonNode) AddFunds(ctx context.Context, addr address.Address, amo
 }
 
 // EnsureFunds adds funds to the given actor in the storage market state to insure it has at least the given amount
-func (n *FakeCommonNode) EnsureFunds(ctx context.Context, addr, wallet address.Address, amount abi.TokenAmount) error {
+func (n *FakeCommonNode) EnsureFunds(ctx context.Context, addr, wallet address.Address, amount abi.TokenAmount, tok shared.TipSetToken) error {
 	if n.EnsureFundsError == nil {
 		balance := n.SMState.Balance(addr)
 		if balance.Available.LessThan(amount) {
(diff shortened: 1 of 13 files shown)


storagemarket/testnodes/testnodes.go
```diff
@@ -180,8 +180,8 @@ func (n *FakeClientNode) OnDealSectorCommitted(ctx context.Context, provider add
 }
 
 // ValidateAskSignature returns the stubbed validation error
```
shannonwells: fix comment

storagemarket/types.go
```diff
@@ -321,7 +321,7 @@ type StorageProviderNode interface {
 
 	OnDealSectorCommitted(ctx context.Context, provider address.Address, dealID abi.DealID, cb DealSectorCommittedCallback) error
 
-	LocatePieceForDealWithinSector(ctx context.Context, dealID abi.DealID) (sectorID uint64, offset uint64, length uint64, err error)
+	LocatePieceForDealWithinSector(ctx context.Context, dealID abi.DealID, tok shared.TipSetToken) (sectorID uint64, offset uint64, length uint64, err error)
```
hannahhoward: Is this method actually talking to the chain? I believe it's not.. though I could be wrong.

laser: Yep. The go-filecoin implementation of this method requires both storage miner actor and storage miner actor states.

laser:
@hannahhoward 

> For the forward facing APIs on Client and Provider, I would prefer not to augment them with a tipset token, and instead have them internally call GetMoreRecentStateID internally

That makes sense to me. I'll make that update tomorrow.

laser:
> Also, should LocatePieceForDealWithinSector take a tipset token? My understanding is this doesn't really talk to the chain so much as it uses the storage miner code directly. This is non-blocking if you can confirm my understanding is wrong.

Yep, [indeed it does](https://github.com/filecoin-project/go-filecoin/blob/7748bab472a148decb704178b03cde6516a305dd/internal/app/go-filecoin/connectors/storage_market/provider.go#L140). At least in go-filecoin.

storagemarket/testnodes/testnodes.go
laser: done!

PR #181, opened 2020-04-03
use PieceCID if provided in QueryParams

### LINKED ISSUE #180: Retrieval Query's support specifying optional PieceCID
In addition to a payload CID, a retrieval query should include an OPTIONAL pieceCID in the params. If present, we should respond in the affirmative ONLY if we have that PieceCID & the CID maps to that PieceCID in the piece store

Rationale: I want to know I'm getting the payloadCID directly from a specific piece, possibly one I know has all of the data I want, and maybe is stored in a specific format (there are plans to evolve piece CID over time)

shannonwells:
Closes #180 

If QueryParams provides a PieceCID, use it and return whether the item is available.

### DIFF
--- a/retrievalmarket/impl/provider.go
+++ b/retrievalmarket/impl/provider.go
@@ -3,7 +3,6 @@ package retrievalimpl
 import (
 	"context"
 	"errors"
-	"fmt"
 	"reflect"
 	"sync"
 
@@ -189,16 +188,17 @@ func (p *provider) HandleQueryStream(stream rmnet.RetrievalQueryStream) {
 	} else {
 		answer.PaymentAddress = paymentAddress
 
-		pieceInfo, err := getPieceInfoFromCid(p.pieceStore, query.PayloadCID, query.PieceCID)
+		pieceCID := cid.Undef
+		if query.PieceCID != nil {
+			pieceCID = *query.PieceCID
+		}
+		pieceInfo, err := getPieceInfoFromCid(p.pieceStore, query.PayloadCID, pieceCID)
 
 		if err == nil && len(pieceInfo.Deals) > 0 {
 			answer.Status = retrievalmarket.QueryResponseAvailable
 			// TODO: get price, look for already unsealed ref to reduce work
 			answer.Size = uint64(pieceInfo.Deals[0].Length) // TODO: verify on intermediate
-
-			if query.PieceCID != cid.Undef {
-				answer.PieceCIDFound = retrievalmarket.QueryItemAvailable
-			}
+			answer.PieceCIDFound = retrievalmarket.QueryItemAvailable
 		}
 
 		if err != nil && !xerrors.Is(err, retrievalmarket.ErrNotFound) {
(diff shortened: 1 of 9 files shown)


retrievalmarket/impl/provider.go
```diff
@@ -287,18 +296,23 @@ func (p *provider) GetPieceSize(c cid.Cid) (uint64, error) {
...
 		pieceInfo, err := pieceStore.GetPieceInfo(pieceBlockLocation.PieceCID)
 		if err == nil {
-			return pieceInfo, nil
+			if pieceCID == cid.Undef || pieceInfo.PieceCID == pieceCID {
+				return pieceInfo, nil
+			}
 		}
 		lastErr = err
 	}
+	if lastErr == nil {
```
shannonwells, line 314: this happens if a payloadCID was found but pieceCID wasn't

retrievalmarket/impl/provider.go
```diff
@@ -277,7 +285,7 @@ func (p *provider) NextBlock(ctx context.Context, id retrievalmarket.ProviderDea
 }
 
 func (p *provider) GetPieceSize(c cid.Cid) (uint64, error) {
-	pieceInfo, err := getPieceInfoFromCid(p.pieceStore, c)
+	pieceInfo, err := getPieceInfoFromCid(p.pieceStore, c, cid.Undef)
```
shannonwells, line 288: This is called by ProposeDeal and DealProposal doesn't know anything about individual PieceCIDs. I'm assuming that client proposing a deal will always do a Query with the PieceCID first so unless PieceCID is added to DealProposal.Params,  we assume client has already determined the provider has what they want.

retrievalmarket/types.go
```diff
@@ -397,8 +397,8 @@ func NewQueryV0(payloadCID cid.Cid) Query {
 
 // QueryResponse is a miners response to a given retrieval query
 type QueryResponse struct {
-	Status QueryResponseStatus
-	//PayloadCIDFound QueryItemStatus // V1 - if a PayloadCid was requested, the result
+	Status        QueryResponseStatus
+	PieceCIDFound QueryItemStatus // V1 - if a PieceCID was requested, the result
```
shannonwells, line 401: this seemed to be a typo/leftover from the old version when Piece/Payload CID weren't distinct

retrievalmarket/types.go
```diff
@@ -373,7 +373,7 @@ const (
 // client is interested in, as well as specific parameters the client is seeking
 // for the retrieval deal
 type QueryParams struct {
-	//PayloadCID                 cid.Cid   // optional, query if miner has this cid in this piece. some miners may not be able to respond.
+	PieceCID cid.Cid // optional, query if miner has this cid in this piece. some miners may not be able to respond.
```
shannonwells, line 376: this seemed to be a typo/leftover from the old version when Piece/Payload CID weren't distinct

retrievalmarket/impl/provider.go
```diff
@@ -287,18 +295,23 @@ func (p *provider) GetPieceSize(c cid.Cid) (uint64, error) {
...
 		if err == nil {
-			return pieceInfo, nil
+			if pieceCID == cid.Undef || pieceInfo.PieceCID == pieceCID {
+				return pieceInfo, nil
+			}
 		}
 		lastErr = err
 	}
+	if lastErr == nil {
+		lastErr = fmt.Errorf("unknown pieceCID %s", pieceCID.String())
```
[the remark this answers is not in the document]
shannonwells: thanks I forgot we use that here

retrievalmarket/types.go
shannonwells: ah, right that was it.

PR #182, opened 2020-04-04
For a storage deal proposal that is rejected, include rejection reason in client response

### LINKED ISSUE #176: StorageMarket: Client can see providers error message
- When a deal is rejected, provider includes message about why it failed in the response.
- client should incorporate the Message property in network.Response into it's own Message property on ClientDeal

Acceptance:
- ReadDealResponse returns a rejection state, plus a Message in Response
- After client errors, Message property of ClientDeal should include Message from Response (but does not need to be limited to that)

shannonwells:
Closes #176

### DIFF
--- a/storagemarket/impl/clientstates/client_fsm.go
+++ b/storagemarket/impl/clientstates/client_fsm.go
@@ -55,8 +55,8 @@ var ClientEvents = fsm.Events{
 		}),
 	fsm.Event(storagemarket.ClientEventDealRejected).
 		From(storagemarket.StorageDealValidating).To(storagemarket.StorageDealFailing).
-		Action(func(deal *storagemarket.ClientDeal, state storagemarket.StorageDealStatus) error {
-			deal.Message = xerrors.Errorf("deal wasn't accepted (State=%d)", state).Error()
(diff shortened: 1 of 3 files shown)


storagemarket/impl/clientstates/client_fsm.go
```diff
@@ -55,8 +55,8 @@ var ClientEvents = fsm.Events{
 		}),
 	fsm.Event(storagemarket.ClientEventDealRejected).
 		From(storagemarket.StorageDealValidating).To(storagemarket.StorageDealFailing).
-		Action(func(deal *storagemarket.ClientDeal, state storagemarket.StorageDealStatus) error {
-			deal.Message = xerrors.Errorf("deal wasn't accepted (State=%d)", state).Error()
+		Action(func(deal *storagemarket.ClientDeal, state storagemarket.StorageDealStatus, reason string) error {
+			deal.Message = xerrors.Errorf("deal failed: (State=%d) %s", state, reason).Error()
```
shannonwells, line 59: Changed this because "deal rejected" is already in the message and the status itself is failed

PR #183, opened 2020-04-06
Improve offline deals

arajasek:
With these changes, the offline deal workflow succeeds

### DIFF
--- a/storagemarket/types.go
+++ b/storagemarket/types.go
@@ -395,8 +395,8 @@ type DataRef struct {
 	TransferType string
 	Root         cid.Cid
 
-	PieceCid  *cid.Cid // Optional, will be recomputed from the data if not given
-	PieceSize abi.UnpaddedPieceSize
+	PieceCid  *cid.Cid              // Optional for non-manual transfer, will be recomputed from the data if not given
+	PieceSize abi.UnpaddedPieceSize // Optional for non-manual transfer, will be recomputed from the data if not given
 }
 
 // The interface provided by the module to the outside world for storage clients.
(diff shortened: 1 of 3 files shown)


storagemarket/types.go
```diff
@@ -426,6 +426,9 @@ type StorageClient interface {
 	// ProposeStorageDeal initiates deal negotiation with a Storage Provider
 	ProposeStorageDeal(ctx context.Context, addr address.Address, info *StorageProviderInfo, data *DataRef, startEpoch abi.ChainEpoch, endEpoch abi.ChainEpoch, price abi.TokenAmount, collateral abi.TokenAmount, rt abi.RegisteredProof) (*ProposeStorageDealResult, error)
 
+	// CalculateCommP calculates the piece CID for a given CAR file and miner
```
whyrusleeping, line 429: comment seems slightly inaccurate

storagemarket/impl/client.go
```diff
@@ -199,9 +200,21 @@ func (c *Client) ProposeStorageDeal(
...
 ) (*storagemarket.ProposeStorageDealResult, error) {
-	commP, pieceSize, err := clientutils.CommP(ctx, c.pio, rt, data)
-	if err != nil {
-		return nil, xerrors.Errorf("computing commP failed: %w", err)
+
+	var commP cid.Cid
+	var pieceSize abi.UnpaddedPieceSize
+	var err error
+
+	if data.TransferType != storagemarket.TTManual {
```
whyrusleeping, line 208: the clientutils.CommP function already checks if its a manual transfer and short circuits. Maybe move this logic down into that?

hannahhoward: yea that makes sense

arajasek:
Thanks for the suggestions, @hannahhoward, I didn't realise after some edits that I wound up just calling a public static method! I've switched to doing it directly from Lotus, I think it's nicer :) 

All suggested changes implemented :)

PR #188, opened 2020-04-08
Add a max piece size to storage asks

arajasek:
We should probably put something in that ensures that the new max piece size is less than or equal to the sector size.

This PR also includes a commit that makes the Storage Market client refuse to propose a deal if piece size > sector size.

### DIFF
--- a/storagemarket/impl/client.go
+++ b/storagemarket/impl/client.go
@@ -204,6 +204,10 @@ func (c *Client) ProposeStorageDeal(
 		return nil, xerrors.Errorf("computing commP failed: %w", err)
 	}
 
+	if uint64(pieceSize.Padded()) > info.SectorSize {
+		return nil, xerrors.New("Cannot propose a deal whose piece size is greater than sector size")
+	}
+
 	dealProposal := market.DealProposal{
 		PieceCID:             commP,
 		PieceSize:            pieceSize.Padded(),
--- a/storagemarket/impl/client.go
+++ b/storagemarket/impl/client.go
@@ -2,6 +2,7 @@ package storageimpl
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/filecoin-project/go-address"
 	cborutil "github.com/filecoin-project/go-cbor-util"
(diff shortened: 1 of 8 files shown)


storagemarket/impl/provider.go
```diff
@@ -262,8 +262,8 @@ func (p *Provider) ListLocalDeals() ([]storagemarket.MinerDeal, error) {
 	return out, nil
 }
 
-func (p *Provider) AddAsk(price abi.TokenAmount, duration abi.ChainEpoch) error {
-	return p.storedAsk.AddAsk(price, duration)
+func (p *Provider) AddAsk(price abi.TokenAmount, duration abi.ChainEpoch, bounds ...abi.PaddedPieceSize) error {
```
hannahhoward, line 265: So I actually had something slightly different in mind when I mentioned variadic arguments -- I was talking about the functional options pattern:
https://www.sohamkamani.com/golang/options-pattern/

I would define the function this way:

```golang
type StorageAskOption func(*storagemarket.StorageAsk)
func (p *Provider) AddAsk(price abi.TokenAmount, duration abi.ChainEpoch, options ...StorageAskOption)
```

and then you can define:

```golang
func MinPieceSize(minPieceSize abi.PaddedPieceSize) StorageAskOptions {
   return func(sa *storagemarket.StorageAsk) {
      sa.MinPieceSize = minPieceSize
   }
}
```

And then, in storedAsk.AddAsk, you can just do:

```golang
ask := &storagemarket.StorageAsk{
		Price:        price,
		Timestamp:    height,
		Expiry:       height + duration,
		Miner:        s.actor,
		SeqNo:        seqno,
		MinPieceSize: defaultMinPieceSize,
                  MaxPieceSize: defaultMaxPieceSize,
}
for _, option := range options {
   option(ask)
}
```

storagemarket/impl/client.go
```diff
@@ -204,6 +205,10 @@ func (c *Client) ProposeStorageDeal(
 		return nil, xerrors.Errorf("computing commP failed: %w", err)
 	}
 
+	if uint64(pieceSize.Padded()) > info.SectorSize {
```
hannahhoward, line 208: Great idea to do this check early -- even if the provider would reject the deal eventually we should go ahead and error if we know the deal won't be accepted

arajasek:
Ah, my  bad, I definitely didn't know what functional options meant! Thanks for the detailed explanation, I've implemented it (and like the pattern, it's much better than what I had)!

PR #197, opened 2020-04-21
add multiple peers per payloadCID

### LINKED ISSUE #101: Track multiple Retrieval Peers
When I store the same piece twice, I should track all miners I've stored it with so I can retrieve from all those miners.

See: https://github.com/filecoin-project/go-fil-markets/blob/master/retrievalmarket/discovery/local.go#L25

Scope:
- local discovery cache is currently pieceCid -> retrievalPeer
- probably want to make it pieceCid -> retrievalPeerList
where retrievalPeerList =
type retrievalPeerList struct {
peers []retrievalPeer
}
(make sure to cbor gen with cbor-gen-for)
- make sure to read current value and check for uniqueness before appending to list

shannonwells:
Closes #101 

Amend storage record of payloadCIDs to  store a list of retrieval peers instead of just one.

### DIFF
--- a/retrievalmarket/discovery/local.go
+++ b/retrievalmarket/discovery/local.go
@@ -6,12 +6,9 @@ import (
 	"github.com/ipfs/go-datastore/namespace"
 	dshelp "github.com/ipfs/go-ipfs-ds-help"
 	cbor "github.com/ipfs/go-ipld-cbor"
-	logging "github.com/ipfs/go-log/v2"
 
 	"github.com/filecoin-project/go-fil-markets/retrievalmarket"
 )
-
-var log = logging.Logger("ret-discovery")
 
 type Local struct {
 	ds datastore.Datastore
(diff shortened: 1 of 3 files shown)


retrievalmarket/discovery/local.go
```diff
@@ -22,17 +19,48 @@ func NewLocal(ds datastore.Batching) *Local {
...
+		newRecord, err = cbor.DumpObject(peerList)
+		if err != nil {
+			return err
+		}
+	}
+
+	return l.ds.Put(key, newRecord)
+}
+
+func hasPeer(peerList []retrievalmarket.RetrievalPeer, peer retrievalmarket.RetrievalPeer) bool {
```
shannonwells, line 57: I'm assuming these are going to be relatively short lists.

hannahhoward: yea shouldn't be too bad

PR #200, opened 2020-04-22
Add a test rule to Makefile, use in CI config

### LINKED ISSUE #199: Add "test" rule to Makefile
As an occasional contributor I found it difficult to work out what was required before putting up a PR. I had to go diving into the CircleCI config. 

Add a `make test` rule (and call it from the Circle config to avoid duplication).

shannonwells:
closes #199 

In the Makefile, set up a test target and make the CI config use it instead what it's doing now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,26 +65,10 @@ jobs:
       executor:
         type: executor
         default: golang
-      go-test-flags:
-        type: string
-        default: "-timeout 5m"
-        description: Flags passed to go test.
-      packages:
-        type: string
-        default: "./..."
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -65,26 +65,10 @@ jobs:
...
-      packages:
-        type: string
-        default: "./..."
-        description: Import paths of packages to be tested.
       test-suite-name:
         type: string
         default: unit
         description: Test suite name to report to CircleCI.
-      gotestsum-format:
-        type: string
```
shannonwells, line 81: short format is the default, so I omitted it in the test target.

PR #202, opened 2020-04-23
Storagemarket/provider allows subscription to events

### LINKED ISSUE #77: User of storage provider can listen for events on deals
As a user of the Storage Provider module, I should be able to listen for updates on deals being processed, for the purposes of logging, or analysis, or any other reason

---

In:
- Add a subscriber / listener function to the Storage Provider
- Pass updates to the listener

---

Acceptance criteria:
- Subscribe to updates on provider
- Propose a deal to a storage provider with a client
- Verify updates come through on listern

shannonwells:
### WAT
closes #77 

### HAO 
* Allow consumers to subscribe to StorageProvider events.
* Use @hannahhoward  new generic threadsafe pubsub for subscription management

### Questions
I'm sending the entire deal to the subscriber. Not sure if that's what we really want though.

### DIFF
--- a/storagemarket/impl/client.go
+++ b/storagemarket/impl/client.go
@@ -42,7 +42,6 @@ type Client struct {
 	// implementation, there's no validation or events on the client side
 	dataTransfer datatransfer.Manager
 	bs           blockstore.Blockstore
-	//fs           filestore.FileStore
 	pio          pieceio.PieceIO
 	discovery    *discovery.Local
 
--- a/storagemarket/impl/client.go
+++ b/storagemarket/impl/client.go
@@ -17,7 +17,6 @@ import (
 	logging "github.com/ipfs/go-log/v2"
 	"golang.org/x/xerrors"
 
-	"github.com/filecoin-project/go-fil-markets/filestore"
 	"github.com/filecoin-project/go-fil-markets/pieceio"
 	"github.com/filecoin-project/go-fil-markets/pieceio/cario"
 	"github.com/filecoin-project/go-fil-markets/retrievalmarket"
@@ -43,7 +42,7 @@ type Client struct {
 	// implementation, there's no validation or events on the client side
 	dataTransfer datatransfer.Manager
 	bs           blockstore.Blockstore
-	fs           filestore.FileStore
+	//fs           filestore.FileStore
 	pio          pieceio.PieceIO
 	discovery    *discovery.Local
 
--- a/storagemarket/impl/provider.go
+++ b/storagemarket/impl/provider.go
@@ -320,7 +320,6 @@ func (p *Provider) dispatch(eventName fsm.EventName, deal fsm.StateType) {
 	}
 	pubSubEvt := internalEvent{evt, &realDeal}
 
-	// because pubSub uses a thread lock
 	go func() {
 		if err := p.pubSub.Publish(pubSubEvt); err != nil {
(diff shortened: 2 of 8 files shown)


storagemarket/impl/provider.go
```diff
@@ -298,6 +303,54 @@ func (p *Provider) UniversalRetrievalEnabled() bool {
...
+		return xerrors.New("wrong type of event")
+	}
+	log.Infof("dispatcher called with valid evt %d", ie.evt)
+	cb(ie.evt, ie.deal)
+	return nil
+}
+
+// -------
+// providerDealEnvironment
+// -------
```
shannonwells, line 352: I kept paging too far down looking at stuff, so I figured a comment might help

storagemarket/impl/provider.go
```diff
@@ -298,6 +303,53 @@ func (p *Provider) UniversalRetrievalEnabled() bool {
...
+	if !ok {
+		log.Errorf("dropped bad event %s", eventName)
+	}
+	realDeal, ok := deal.(storagemarket.MinerDeal)
+	if !ok {
+		log.Errorf("not a deal %v", deal)
+	}
+	pubSubEvt := internalEvent{evt, &realDeal}
+
+	go func() {
```
shannonwells, line 323: Without this the integration test fails when there is a subscriber, so it seems pubsub is causing a deadlock

storagemarket/impl/client.go
```diff
@@ -43,7 +42,7 @@ type Client struct {
 	// implementation, there's no validation or events on the client side
 	dataTransfer datatransfer.Manager
 	bs           blockstore.Blockstore
-	fs           filestore.FileStore
+	//fs           filestore.FileStore
```
hannahhoward, line 45: are we really not using this? If so just delete it.

storagemarket/types.go
```diff
@@ -277,6 +277,9 @@ type StorageDeal struct {
 	market.DealState
 }
 
+// Subscriber is a callback that is called when events are emitted
+type ProviderSubscriber func(event ProviderEvent, deal *MinerDeal)
```
hannahhoward, line 281: this should not be a pointer. You cannot modify the miner deal, so pass the value.

storagemarket/impl/provider.go
hannahhoward: this is not correct. Remove the go routine, add a buffer one on the dealchan.

It locks because the send to the dealChan locks in the integration test. Channel writes block without a buffer on the channel.

hannahhoward: that alone will probably not fix it, cause you are writing to the deal chan each time you update state. you either need to make sure you're seperately clearing the deal chan every time, or you need to make sure to write to it only on certain states and then clear it. the listener should NOT be a blocking operation.

storagemarket/integration_test.go
```diff
@@ -115,16 +122,48 @@ func TestMakeDeal(t *testing.T) {
...
+	for seenDeal.State != storagemarket.StorageDealCompleted {
+		select {
+		case seenDeal = <-dealChan:
+			actualStates = append(actualStates, seenDeal.State)
+		case <-ctx.Done():
+			t.Fatalf("never saw event")
+		}
+	}
+
+	expectedStates := []storagemarket.StorageDealStatus{
```
hannahhoward, line 138: Love this check!

PR #223, opened 2020-05-05
Storage Market Changes Based On Lotus Integration

hannahhoward:
# Goals

Get the latest go-fil-markets integrated with lotus so we can cut a 0.2.0 branch

# Implementation

The integration revealed we can't as it turns out, wait with confidence on a chain message, only a state change -- we could technically call StateGetReceipt or StateSearchMessage -- but this would likely be extremely slow :(

So, in the absence, I removed confidence for now. 

Also, our callback takes an exit code and data, but does not account for the possibility that the wait call itself errors (i.e. message was never posted). So I added third error parameter to the callback to capture this.

I also added a context to the call (LOTUS and I imagine GFC API expect a context), and made a few dependency updates

### DIFF
--- a/storagemarket/impl/providerstates/provider_states.go
+++ b/storagemarket/impl/providerstates/provider_states.go
@@ -192,11 +192,14 @@ func EnsureProviderFunds(ctx fsm.Context, environment ProviderDealEnvironment, d
 func WaitForFunding(ctx fsm.Context, environment ProviderDealEnvironment, deal storagemarket.MinerDeal) error {
 	node := environment.Node()
 
-	return node.WaitForMessage(deal.AddFundsCid, storagemarket.ChainConfidence, func(code exitcode.ExitCode, bytes []byte) error {
-		if code == exitcode.Ok {
-			return ctx.Trigger(storagemarket.ProviderEventFunded)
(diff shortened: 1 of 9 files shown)


storagemarket/impl/providerstates/provider_states.go
```diff
@@ -220,32 +223,35 @@ func PublishDeal(ctx fsm.Context, environment ProviderDealEnvironment, deal stor
...
+			Proposal:       deal.ProposalCid,
+			PublishMessage: &deal.PublishCid,
+		})
+
+		if err != nil {
+			return ctx.Trigger(storagemarket.ProviderEventSendResponseFailed, err)
 		}
 
-		return ctx.Trigger(storagemarket.ProviderEventDealPublishError, xerrors.Errorf("PublishStorageDeals exit code: %w", code))
+		if err := environment.Disconnect(deal.ProposalCid); err != nil {
```
shannonwells, line 249: I'm curious about Disconnect. Why is Disconnect called only when is no error?  Also, there is a Disconnect call in SendSignedResponse, although in the tests it shows that it's harmless to do it twice.

hannahhoward: Disconnect gets called in the DealFailing handler, so no need to call it here. It probably should not be calling disconnect in SendSignedResponse and I will check that and add a ticket if needed.

Issue #192, opened 2020-04-15
Storage Provider sends tentative acceptance


hannahhoward:
When a storage provider validates a deal, it should send a tentative acceptance. We can probably use the existing network response and adjust the statuses. 

**My exact suggested flow is as follows:**
- ValidateDeal on provider, move status to StorageDealWaitingForData and send response with status to client.
- For normal deal, client initiates data transfer and moves its own status to storagedealtransferring
- It should ALSO listen for the complete on the data transfer before it moves on to the normal read of storage deal proposal accepted
- For manual deal, client does nothing. It can move right on to the read for proposal accepted (this is problematic too but let's not address here)
- On provider, listen to data transfer events of both open AND complete. Open should move our status from StorageDealWaitingForData to StorageDealTransferring. When the request completes, we move onto StorageDealVerifyData.
- Manual DT for the provder is unchanged. We can probably remove the use of StorageDealProposalAccepted right after ValidateDealProposal and the TransferData handler entirely

PR #229, opened 2020-05-06
Round-trip storage/retrieval test

### LINKED ISSUE #74: Roundtrip Storage/Retrieval Market Integration Test
We should have a least a happy path for testing that a client can:
- make a storage deal with a provider
- storage provider receives and completes storing deal
- client can make a retrieval deal for their data
- provider responds and sends back data stored

shannonwells:
## Motivation
We need a storage-retrieval round trip integration test.  Closes #74 

## Change Summary:
* make a test that combines the storage and retrieval market integration tests
* move TestPeerResolver to shared testutils so it can be used more widely
* make everybody using `LoadUnixFSFile` have to specify the path to their fixture file, starting from package root, so it can be used more easily by different packages.
* export AllSelector
* please the linter with a bunch of nitpicky changes

More detail about the integration test:
This is essentially the combined storage and retrieval integration tests, with most assertions removed and the remaining ones converted to requires.  For retrieval code, the integration test code is converted to use the test harness method in storage integration test, in particular so the LibP2PTestData can be shared from the storage to the retrieval harness, as well as other things such as the Payload CID, so that the stored file can be retrieved.

### DIFF
--- a/retrievalmarket/impl/common.go
+++ b/retrievalmarket/impl/common.go
@@ -7,7 +7,7 @@ import (
 	"github.com/ipld/go-ipld-prime/traversal/selector/builder"
 )
 
-func allSelector() ipld.Node {
+func AllSelector() ipld.Node {
 	ssb := builder.NewSelectorSpecBuilder(basicnode.Style.Any)
 	return ssb.ExploreRecursive(selector.RecursionLimitNone(),
 		ssb.ExploreAll(ssb.ExploreRecursiveEdge())).
--- a/shared_testutil/mocknet.go
+++ b/shared_testutil/mocknet.go
@@ -28,9 +28,6 @@ import (
 	"github.com/ipfs/go-unixfs/importer/helpers"
 	"github.com/ipld/go-ipld-prime"
 	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
-	basicnode "github.com/ipld/go-ipld-prime/node/basic"
-	"github.com/ipld/go-ipld-prime/traversal/selector"
-	"github.com/ipld/go-ipld-prime/traversal/selector/builder"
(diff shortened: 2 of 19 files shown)


retrievalmarket/impl/common.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/ipld/go-ipld-prime/traversal/selector/builder"
 )
 
-func allSelector() ipld.Node {
+func AllSelector() ipld.Node {
```
ingar, line 10: Seems like while you are exporting this, this is code snippet is used all over the place in tests and in the storage market...

shared_testutil/mocknet.go
```diff
@@ -151,13 +153,13 @@ const unixfsLinksPerLevel = 1024
 // LoadUnixFSFile injects the fixture `filename` into the given blockstore from the
 // fixtures directory. If useSecondNode is true, fixture is injected to the second node;
 // otherwise the first node gets it
-func (ltd *Libp2pTestData) LoadUnixFSFile(t *testing.T, filename string, useSecondNode bool) ipld.Link {
+func (ltd *Libp2pTestData) LoadUnixFSFile(t *testing.T, fixturesPath string, useSecondNode bool) ipld.Link {
 
 	// read in a fixture file
-	path, err := filepath.Abs(filepath.Join("fixtures", filename))
+	fpath, err := filepath.Abs(filepath.Join(thisDir(t), "..",fixturesPath))
```
ingar, line 159: The linter allowed this??  heh.

ingar (commented):
Looks good, probably there will be some code extraction once #192 lands...

ingar (approved):
Thanks

PR #230, opened 2020-05-08
Don't Keep Streams Open

hannahhoward:
# Goals

Our approach has been to just leave streams open, and read form them at will in storage market, and this has caused several issues.

This PR moves to closing streams as soon as we send information in the storage market. Retrieval will move to this also, but only after the data transfer move has happened

# Implementation

- Change to write and reads of deal proposals/responses opening the stream, sending the data, and closing the stream
- Only read responses in response handlers
- Fix a number of CBOR gen issues (will be instituting autocheck shortly)
- Remove use of conmanager ( need to delete files)
- Update with waiting for response state.
- move network types to root storage market types file so response can live on client deal

# Follow-up

- remove conn manager
- institute nil types


storagemarket/impl/providerstates/provider_states_test.go
```diff
@@ -305,7 +304,7 @@ func TestEnsureProviderFunds(t *testing.T) {
 			},
 			dealInspector: func(t *testing.T, deal storagemarket.MinerDeal) {
 				require.Equal(t, storagemarket.StorageDealProviderFunding, deal.State)
-				require.Equal(t, cids[0], deal.AddFundsCid)
+				require.Equal(t, &cids[0], deal.AddFundsCid)
```
ingar, line 307: I am really confused about pointers, equality and how this works with serialization into and out of the store. I guess maybe in this case `require.Equal` is dereferencing the pointers under the hood?

I guess a more general question, why do we have a mix of `*cid.Cid` and `cid.Cid` in the `MinerDeal` and `ClientDeal`?  Should we be using one or the other?

storagemarket/impl/provider.go
```diff
@@ -139,9 +139,8 @@ func (p *Provider) HandleDealStream(s network.StorageDealStream) {
 	err := p.receiveDeal(s)
 	if err != nil {
 		log.Errorf("%+v", err)
-		s.Close()
-		return
 	}
+	s.Close()
 }
 
 func (p *Provider) receiveDeal(s network.StorageDealStream) error {
```
shannonwells, line 146: I  know it was in use before these changes, but I don't think any real purpose is being served by the receiveDeal abstraction. It's short and used only in HandleDealStream.

storagemarket/impl/providerstates/provider_states_test.go
hannahhoward: yes -- good point, non-obvious, but this is knowledge that should be shared by everyone. A CID that is not ALWAYS set should be a pointer. If a data structure has a regular cid with a value of cid.Undef, it will NOT serialize to CBOR. These changes are actually corrections to already merged code that had to be made once I ran our cbor generation scripts. I'm working. @shannonwells is working on a ticket to add checking that CBOR generation has been run to our backlog.

PR #244, opened 2020-05-11
Tentative acceptance protocol

### LINKED ISSUE #192: Storage Provider sends tentative acceptance

ingar:
## Summary
Switch piece data transfer to a client push.

resolves #192

### DIFF
--- a/shared_testutil/testutil.go
+++ b/shared_testutil/testutil.go
@@ -11,7 +11,10 @@ import (
 	blocksutil "github.com/ipfs/go-ipfs-blocksutil"
 	"github.com/jbenet/go-random"
 	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
+	"github.com/filecoin-project/go-fil-markets/storagemarket"
 )
 
 var blockGenerator = blocksutil.NewBlockGenerator()
@@ -93,4 +96,13 @@ func TestVoucherEquality(t *testing.T, a, b *paych.SignedVoucher) {
 	bB, err := cborutil.Dump(b)
 	require.NoError(t, err)
 	require.True(t, bytes.Equal(aB, bB))
+}
+
+// AssertDealState asserts equality of StorageDealStatus but with better error messaging
+func AssertDealState(t *testing.T, expected storagemarket.StorageDealStatus, actual storagemarket.StorageDealStatus) {
+	assert.Equal(t, expected, actual,
+		"Unexpected deal status\nexpected: %s (%d)\nactual  : %s (%d)",
+		storagemarket.DealStates[expected], expected,
+		storagemarket.DealStates[actual], actual,
+	)
 }
--- a/storagemarket/impl/client.go
+++ b/storagemarket/impl/client.go
@@ -17,6 +17,7 @@ import (
 	"github.com/ipfs/go-datastore"
 	blockstore "github.com/ipfs/go-ipfs-blockstore"
 	logging "github.com/ipfs/go-log/v2"
+	"github.com/ipld/go-ipld-prime"
 	"github.com/libp2p/go-libp2p-core/peer"
 	"golang.org/x/xerrors"
 
@@ -29,6 +30,7 @@ import (
 	"github.com/filecoin-project/go-fil-markets/storagemarket/impl/clientstates"
 	"github.com/filecoin-project/go-fil-markets/storagemarket/impl/clientutils"
 	"github.com/filecoin-project/go-fil-markets/storagemarket/impl/connmanager"
+	"github.com/filecoin-project/go-fil-markets/storagemarket/impl/dtutils"
 	"github.com/filecoin-project/go-fil-markets/storagemarket/network"
 )
 
@@ -39,11 +41,6 @@ var _ storagemarket.StorageClient = &Client{}
 type Client struct {
 	net network.StorageMarketNetwork
 
-	// dataTransfer
-	// TODO: once the data transfer module is complete, the
-	// client will listen to events on the data transfer module
-	// Because we are using only a fake DAGService
-	// implementation, there's no validation or events on the client side
 	dataTransfer datatransfer.Manager
 	bs           blockstore.Blockstore
 	pio          pieceio.PieceIO
@@ -89,6 +86,10 @@ func NewClient(
 		return nil, err
 	}
 	c.statemachines = statemachines
+
+	// register a data transfer event handler -- this will send events to the state machines based on DT events
+	dataTransfer.SubscribeToEvents(dtutils.ClientDataTransferSubscriber(statemachines))
+
 	return c, nil
 }
 
@@ -385,4 +386,7 @@ func (c *clientDealEnvironment) CloseStream(proposalCid cid.Cid) error {
 	return c.c.conns.Disconnect(proposalCid)
 }
 
-var _ clientstates.ClientDealEnvironment = &clientDealEnvironment{}
+func (c *clientDealEnvironment) StartDataTransfer(ctx context.Context, to peer.ID, voucher datatransfer.Voucher, baseCid cid.Cid, selector ipld.Node) error {
+	_, err := c.c.dataTransfer.OpenPushDataChannel(ctx, to, voucher, baseCid, selector)
+	return err
+}
--- a/storagemarket/impl/clientstates/client_fsm.go
+++ b/storagemarket/impl/clientstates/client_fsm.go
@@ -36,12 +36,6 @@ var ClientEvents = fsm.Events{
 		}),
 	fsm.Event(storagemarket.ClientEventDealProposed).
 		From(storagemarket.StorageDealFundsEnsured).To(storagemarket.StorageDealWaitingForDataRequest),
-	fsm.Event(storagemarket.ClientEventDealStreamLookupErrored).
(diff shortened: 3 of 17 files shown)


shared_testutil/testutil.go
```diff
@@ -94,3 +96,12 @@ func TestVoucherEquality(t *testing.T, a, b *paych.SignedVoucher) {
 	require.NoError(t, err)
 	require.True(t, bytes.Equal(aB, bB))
 }
+
+// AssertDealState asserts equality of StorageDealStatus but with better error messaging
```
hannahhoward, line 100: ❤️  Yay for this

storagemarket/impl/client.go
```diff
@@ -372,4 +369,19 @@ func (c *clientDealEnvironment) CloseStream(proposalCid cid.Cid) error {
...
+func (c *clientDealEnvironment) StartDataTransfer(ctx context.Context, to peer.ID, voucher datatransfer.Voucher, baseCid cid.Cid, selector ipld.Node) error {
+	_, err := c.c.dataTransfer.OpenPushDataChannel(ctx, to, voucher, baseCid, selector)
+	return err
+}
+
+func (c *clientDealEnvironment) OnDataTransferEvent(ctx context.Context, eventCode datatransfer.EventCode, cb providerstates.OnDataTransferEventCallback) datatransfer.Unsubscribe {
+	// Can there be overlapping Open/Complete events?
+	return c.c.dataTransfer.SubscribeToEvents(func(event datatransfer.Event, channelState datatransfer.ChannelState) {
+		if event.Code == eventCode {
+			err := cb(event)
```
hannahhoward, line 381: SubscribeToEvents will get called for every transfer possible. I don't think this is a good way to setup this listener.

You could use the approach above and add a Proposal CID and check the Voucher (also check that it's a data transfer voucher

OR

You could use a single subscriber, check the voucher, and send events in from the outside -- the same way as was used with DataTransferSubscriber on the provider.

I think this is more efficient.

And then you just avoid a StateHandler entirely for StorageDealTransferring.

storagemarket/impl/provider.go
```diff
@@ -392,6 +388,17 @@ func (p *providerDealEnvironment) Ask() storagemarket.StorageAsk {
 	return *sask.Ask
 }
 
+func (p *providerDealEnvironment) OnDataTransferEvent(ctx context.Context, eventCode datatransfer.EventCode, cb providerstates.OnDataTransferEventCallback) datatransfer.Unsubscribe {
```
hannahhoward, line 391: same problem as client. Again, I think this is better done via a single subscriber and reading the voucher to know where to route the event

hannahhoward: it generally should work with the existing DataTransferSubscriber with only a couple modifications

storagemarket/impl/providerstates/provider_states.go
```diff
@@ -101,35 +103,42 @@ func ValidateDealProposal(ctx fsm.Context, environment ProviderDealEnvironment,
...
 
-	return ctx.Trigger(storagemarket.ProviderEventDataTransferInitiated)
+	environment.OnDataTransferEvent(ctx.Context(), datatransfer.Open, func(event datatransfer.Event) error {
+		return ctx.Trigger(storagemarket.ProviderEventDataTransferInitiated)
+	})
+
+	return nil
+}
+
+func TransferringData(ctx fsm.Context, environment ProviderDealEnvironment, deal storagemarket.MinerDeal) error {
```
hannahhoward, line 136: This handler should be unneccesary if you have a general data transfer subscriber

storagemarket/impl/clientstates/client_states.go
```diff
@@ -77,6 +86,50 @@ func ProposeDeal(ctx fsm.Context, environment ClientDealEnvironment, deal storag
...
+	)
+
+	if err != nil {
+		return ctx.Trigger(storagemarket.ProviderEventDataTransferFailed, xerrors.Errorf("failed to open push data channel: %w", err))
+	}
+
+	return ctx.Trigger(storagemarket.ClientEventDataTransferInitiated)
+}
+
+func TransferringData(ctx fsm.Context, environment ClientDealEnvironment, deal storagemarket.ClientDeal) error {
```
hannahhoward, line 126: This should be unneccesary with a single subscriber for the client

storagemarket/impl/client.go
ingar: That makes way more sense.  I was thinking of the "resuming" scenario, but this covers all the cases.

storagemarket/impl/providerstates/provider_states.go
```diff
@@ -45,6 +45,8 @@ type ProviderDealEnvironment interface {
 	DealAcceptanceBuffer() abi.ChainEpoch
 }
 
+type OnDataTransferEventCallback func(event datatransfer.Event) error
```
hannahhoward, line 48: Is this used?

storagemarket/impl/providerstates/provider_fsm.go
```diff
@@ -24,18 +24,18 @@ var ProviderEvents = fsm.Events{
 			deal.Message = xerrors.Errorf("deal rejected: %w", err).Error()
 			return nil
 		}),
-	fsm.Event(storagemarket.ProviderEventDealAccepted).
-		From(storagemarket.StorageDealValidating).To(storagemarket.StorageDealProposalAccepted),
+	fsm.Event(storagemarket.ProviderEventDataRequested).
```
hannahhoward, line 27: Is ProviderEventDataRequested and ProviderEventWaitingForManualData functionally equivalent at this point? Maybe we should just combine them.

I ask cause I noticed that in the provider_states.ValidateDealProposal, at the end there is an if statement dispatching two different events, but no actual different code run other than a log message. I'd just make these a single case. (waiting for the client to send us data over data transfer is essentially the same as waiting for the client to send us a hard drive)

storagemarket/impl/providerstates/provider_states_test.go
```diff
@@ -962,10 +916,19 @@ type fakeEnvironment struct {
 	fs                      filestore.FileStore
 	pieceStore              piecestore.PieceStore
 	dealAcceptanceBuffer    abi.ChainEpoch
+	dtCallbacks             map[datatransfer.EventCode]providerstates.OnDataTransferEventCallback
 	expectedTags            map[string]struct{}
 	receivedTags            map[string]struct{}
 }
 
+func (fe *fakeEnvironment) OnDataTransferEvent(ctx context.Context, eventCode datatransfer.EventCode, cb providerstates.OnDataTransferEventCallback) datatransfer.Unsubscribe {
```
hannahhoward, line 924: delete?

storagemarket/impl/clientstates/client_states_test.go
```diff
@@ -356,9 +419,36 @@ func makeNode(params nodeParams) storagemarket.StorageClientNode {
...
+	fe.startDataTransferCalls = append(fe.startDataTransferCalls, dataTransferParams{
+		to:       to,
+		voucher:  voucher,
+		baseCid:  baseCid,
+		selector: selector,
+	})
+	return nil
+}
+
+func (fe *fakeEnvironment) OnDataTransferEvent(ctx context.Context, eventCode datatransfer.EventCode, cb providerstates.OnDataTransferEventCallback) datatransfer.Unsubscribe {
```
hannahhoward, line 446: delete?

storagemarket/impl/client.go
ingar: Oh right.

storagemarket/impl/providerstates/provider_fsm.go
ingar: 👍

storagemarket/impl/clientstates/client_fsm.go
```diff
@@ -34,25 +34,41 @@ var ClientEvents = fsm.Events{
...
 			return nil
 		}),
+	fsm.Event(storagemarket.ClientEventUnexpectedDealState).
+		From(storagemarket.StorageDealWaitingForDataRequest).To(storagemarket.StorageDealFailing).
+		Action(func(deal *storagemarket.ClientDeal, status storagemarket.StorageDealStatus) error {
+			deal.Message = xerrors.Errorf("unexpected deal status while waiting for data request: %d", status).Error()
+			return nil
+		}),
+	fsm.Event(storagemarket.ClientEventDataTransferFailed).
+		From(storagemarket.StorageDealWaitingForDataRequest).To(storagemarket.StorageDealFailing).
```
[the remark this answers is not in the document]
ingar: I would have never seen this.

PR #257, opened 2020-05-18
Feat/change prefixes 256

### LINKED ISSUE #256: Remove all hardcoded datastore prefixes
Generally, the node should configure the prefix on any datastore usage.

This ticket involves:
- [ ] remove places in code where datastore prefix is marked for this issue (TODOs should be present)
- [ ] cut a new release
- [ ] update both node implementations and provide data store default prefixes (since node maintainers will likely not know enough to set this up right initially.

shannonwells:
# problem
We shouldn't be specifying the root namespace, rather give it a prefix only and allow nodes to add their own, otherwise attach to root.
Part of #256

# solution
delete text specifying base namespace

### DIFF
--- a/retrievalmarket/discovery/local.go
+++ b/retrievalmarket/discovery/local.go
@@ -3,7 +3,6 @@ package discovery
 import (
 	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-datastore"
-	"github.com/ipfs/go-datastore/namespace"
 	dshelp "github.com/ipfs/go-ipfs-ds-help"
 	cbor "github.com/ipfs/go-ipld-cbor"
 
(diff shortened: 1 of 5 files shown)


retrievalmarket/impl/provider.go
```diff
@@ -28,7 +28,7 @@ import (
 )
 
 // ProviderDsPrefix is the datastore for the provider key
-var ProviderDsPrefix = "/retrievals/provider"
+var ProviderDsPrefix = "/provider"
```
hannahhoward, line 31: sorry should be no prefix here

retrievalmarket/impl/client.go
```diff
@@ -46,7 +46,7 @@ type client struct {
 var _ retrievalmarket.RetrievalClient = &client{}
 
 // ClientDsPrefix is the datastore for the client retrievals key
-var ClientDsPrefix = "/retrievals/client"
+var ClientDsPrefix = "/client"
```
hannahhoward, line 49: should be no prefix here (remove namespace wrap)

retrievalmarket/discovery/local.go
```diff
@@ -15,7 +15,7 @@ type Local struct {
 }
 
 func NewLocal(ds datastore.Batching) *Local {
-	return &Local{ds: namespace.Wrap(ds, datastore.NewKey("/deals/local"))}
```
hannahhoward, line 18: should be no prefix here (remove namespace wrap)

PR #263, opened 2020-05-20
Remove storedask from provider

hannahhoward:
# Goals

Add one more element of datastore customization

# Implementation

- Move stored ask out of provider so its setup and configured seperately. 
- Also allow customizing the DS key for stored ask
- Remove prefix for provider in DS store
- Modify integration tests

### DIFF
--- a/retrievalmarket/storage_retrieval_integration_test.go
+++ b/retrievalmarket/storage_retrieval_integration_test.go
@@ -11,12 +11,13 @@ import (
 
 	"github.com/filecoin-project/go-address"
 	datatransfer "github.com/filecoin-project/go-data-transfer"
-	"github.com/filecoin-project/go-data-transfer/impl/graphsync"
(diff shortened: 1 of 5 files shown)


retrievalmarket/storage_retrieval_integration_test.go
```diff
@@ -220,6 +222,8 @@ func newStorageHarness(ctx context.Context, t *testing.T) *storageHarness {
 	)
 	require.NoError(t, err)
 	dt2 := graphsyncimpl.NewGraphSyncDataTransfer(td.Host2, td.GraphSync2, td.DTStoredCounter2)
+	storedAsk, err := storedask.NewStoredAsk(td.Ds2, datastore.NewKey("latest-ask"), providerNode, providerAddr)
+	assert.NoError(t, err)
```
shannonwells, line 226: It's a nitpick but I think this should be a require.NoError since the test doesn't make sense unless this succeeds.

Issue #78, opened 2020-01-28
Storage Client Can Resume Deals On Relaunch


hannahhoward:
As a client, I would like my in progress storage deals to last between restarts, so that they resume when my node comes back online

--

In:

- Storage Deals Are Persisted To Disk On Client As they change state
- On restart, in progress deals resume at their most recent state transition

--

Acceptance Criteria:
- Make a storage deal
- Shut down client
- Restart client
- Verify deal completes assuming no timeout

PR #270, opened 2020-06-05
Restartable storage deals

### LINKED ISSUE #79: Storage Provider can resume deals on relaunch
As a provider, I would like my in progress deals to last between restarts, so that they resume when my node comes back online

--

In:

- Retrieval Deals Are Persisted To Disk On Provider As they change state
- On restart, in progress deals resume at their most recent state transition

--

Acceptance Criteria:
- Start a storage deal
- Shut down provider
- Restart Provider
- Verify deal completes assuming no timeout

ingar:
## Summary
Use the finality states in to declare terminal states for client and provider deals.  On starting up, send either a `ProviderEventRestart` or `ClientEventRestart` event to each deal that isn't in a terminal state.

Resolves #79, #78

## Note
This does not make every deal restartable at any point.  It just adds the plumbing so that the restart event is sent, and will work in some cases.

### Before this is ready to merge
- [x] Merge https://github.com/filecoin-project/go-statemachine/pull/10
- [x] Update `go.mod` to point to new go-statemachine

### DIFF
--- a/storagemarket/impl/client.go
+++ b/storagemarket/impl/client.go
@@ -74,14 +74,11 @@ func NewClient(
 		conns:        connmanager.NewConnManager(),
 	}
 
-	statemachines, err := fsm.New(ds, fsm.Parameters{
-		Environment:     &clientDealEnvironment{c},
-		StateType:       storagemarket.ClientDeal{},
-		StateKeyField:   "State",
-		Events:          clientstates.ClientEvents,
-		StateEntryFuncs: clientstates.ClientStateEntryFuncs,
-		Notifier:        c.dispatch,
-	})
+	statemachines, err := NewClientStateMachine(
+		ds,
+		&clientDealEnvironment{c},
+		c.dispatch,
+	)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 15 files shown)


storagemarket/impl/client.go
```diff
@@ -348,14 +386,19 @@ func (c *clientDealEnvironment) Node() storagemarket.StorageClientNode {
...
-	s, err := c.c.net.NewDealStream(p)
-	if err != nil {
-		return err
-	}
-	err = c.c.conns.AddStream(proposalCid, s)
+	s, err := c.c.conns.DealStream(proposalCid)
+
 	if err != nil {
-		return err
+		s, err = c.c.net.NewDealStream(p)
```
hannahhoward, line 392: small thing but maybe extract these six lines to a function -- you use them up in restart as well.

storagemarket/impl/client.go
```diff
@@ -301,6 +299,34 @@ func (c *Client) SubscribeToEvents(subscriber storagemarket.ClientSubscriber) sh
...
+	if err != nil {
+		return err
+	}
+
+	for _, deal := range deals {
+		if c.statemachines.IsTerminated(deal) {
+			continue
+		}
+
+		s, err := c.net.NewDealStream(deal.Miner)
```
hannahhoward, line 314: So, both ClientDeal & MinerDeal have a value called ConnectionClosed. This is set to true when all the communication over libp2p that happens between provider/miner has happened (basically happens around the time of publishing). If ConnectionClosed is true, you should SKIP opening and adding the stream here.

storagemarket/impl/client.go
```diff
@@ -94,6 +91,7 @@ func NewClient(
 }
 
 func (c *Client) Run(ctx context.Context) {
+	_ = c.restartDeals()
```
hannahhoward, line 94: We should rename this function Start for parity with Provider and also have it return and error so we're not suppressing the error here.

storagemarket/impl/client.go
ingar: Oh, I didn't realize that's what `ConnectionClosed` was, will put in that check.

PR #282, opened 2020-06-18
Query protocol

ingar:
## Problem
We need a network protocol to query the status of a deal.  This will help with a number of things like resuming interrupted deal negotiation, and not having to hold open connections from clients to providers for long periods of time.

## Solution
Add a new protocol which returns the current provider deal state


storagemarket/types.go
```diff
@@ -539,6 +540,17 @@ type DataRef struct {
 	PieceSize abi.UnpaddedPieceSize // Optional for non-manual transfer, will be recomputed from the data if not given
 }
 
+// ProviderDealState represents a Provider's current state of a deal
+type ProviderDealState struct {
+	State       StorageDealStatus
```
hannahhoward, line 545: Is there a reason this is not just a MinerDeal?

ingar: It was originally to deal with the case where the miner deal was not found.  However, now that we have signature verification, we have to look up the miner deal to find the StorageClient's address.  So this struct should flattened.

storagemarket/impl/provider.go
```diff
@@ -398,14 +462,28 @@ func (c *Provider) restartDeals() error {
...
 	return nil
 }
 
+func (p *Provider) sign(ctx context.Context, data interface{}) (*crypto.Signature, error) {
+	tok, _, err := p.spn.GetChainHead(ctx)
+	if err != nil {
+		return nil, xerrors.Errorf("couldn't get chain head: %w", err)
+	}
+
+	sig, err := providerutils.SignMinerData(ctx, data, p.actor, tok, p.spn.GetMinerWorkerAddress, p.spn.SignBytes)
```
hannahhoward, line 479: I believe you can just return this function and skip the remaining lines?

storagemarket/types.go
```diff
@@ -582,6 +588,8 @@ type StorageClient interface {
 	// GetAsk returns the current ask for a storage provider
 	GetAsk(ctx context.Context, info StorageProviderInfo) (*SignedStorageAsk, error)
 
+	GetDealStatus(ctx context.Context, info StorageProviderInfo, proposalCid cid.Cid) (*ProviderDealState, error)
```
hannahhoward, line 591: I feel somewhat strongly this method should be called GetProviderDealState or GetDealStatusFromProvider to disambiguate from GetLocalDeal

ingar: Done

Issue #278, opened 2020-06-16
Storage Client Track Deals To Expiry or Slashing


hannahhoward:
As a Storage Client user, I should be able to see when my active deal becomes either expired or gets slashed, so that I can know it's current status all the way to completion

Tasks and Suggested Architecture:

Market Side:
- Add a client node interface function to wait for a storage deal to get slashed or expire. The method most likely looks like:
(rough sketch)
```golang

type DealExpiredCallback func(err error)
type DealSlashedCallback func(slashEpoch abi.ChainEpoch, err error)
type StorageClientNode {
...
  OnDealExpiredOrSlashed(ctx context.Context, dealID abi.DealID, onDealExpired DealSectorExpiredCallback, onDealSlashed DealSlashedCallback)
}

```

- StorageDealActive is no longer a finality state for StorageClient, add two new finality states to replace it -- StorageDealExpired and StorageDealSlashed (make sure to update StrorageDealStatuses map)
- Add SlashEpoch to ClientDeal
- Add a StateEntryFunc for StorageDealActive that calls OnDealExpiredOrSlashed
   - Dispatch events in callbacks to transition state to expired or slashed
- Record SlashEpoch when deal expires

LOTUS Side:
- Implement the OnDealExpiredOrSlashed callback. Most likely you will want to use the lotus Events API twice, once for Slashed and once for Expired. Slash can be done with the EventsCalled API, similar to how OnDealSectorComitted works, and looking for the OnMinerSectorsTerminate message and checking for one that matches the relevant deal ID. Expired can be handled with the EventsAt API (ChainAt function) passing the deal.EndEpoch as the AT parameter

PR #284, opened 2020-06-18
Storage client - track deals to expiry or slashing

dirkmc:
Implementation of the Market side of issue https://github.com/filecoin-project/go-fil-markets/issues/278:
```
OnDealExpiredOrSlashed(
  ctx context.Context,
  dealID abi.DealID,
  onDealExpired DealExpiredCallback,
  onDealSlashed DealSlashedCallback) error
```

### DIFF
--- a/storagemarket/impl/clientstates/client_fsm.go
+++ b/storagemarket/impl/clientstates/client_fsm.go
@@ -108,6 +108,20 @@ var ClientEvents = fsm.Events{
 		}),
 	fsm.Event(storagemarket.ClientEventDealActivated).
 		From(storagemarket.StorageDealSealing).To(storagemarket.StorageDealActive),
+	fsm.Event(storagemarket.ClientEventDealSlashed).
+		From(storagemarket.StorageDealActive).To(storagemarket.StorageDealSlashed).
+		Action(func(deal *storagemarket.ClientDeal, slashEpoch abi.ChainEpoch) error {
+			deal.SlashEpoch = slashEpoch
+			return nil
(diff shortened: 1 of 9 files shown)


storagemarket/impl/clientstates/client_fsm.go
```diff
@@ -108,6 +108,20 @@ var ClientEvents = fsm.Events{
...
 	fsm.Event(storagemarket.ClientEventDealActivated).
 		From(storagemarket.StorageDealSealing).To(storagemarket.StorageDealActive),
+	fsm.Event(storagemarket.ClientEventDealSlashed).
+		From(storagemarket.StorageDealActive).To(storagemarket.StorageDealSlashed).
+		Action(func(deal *storagemarket.ClientDeal, slashEpoch abi.ChainEpoch) error {
+			deal.SlashEpoch = slashEpoch
+			return nil
+		}),
+	fsm.Event(storagemarket.ClientEventDealExpired).
+		From(storagemarket.StorageDealActive).To(storagemarket.StorageDealExpired),
```
ingar, line 118: I don't know if it matters, or if we can deduce this from the original deal proposal, but is it useful to know when the deal expired?

hannahhoward: Should be determined from dealProposal.EndEpoch

dirkmc:
> I can't tell if we should be proud the code is in a place where a new person can be productive quickly

The code is really well-structured and easy to follow 😍

PR #297, opened 2020-06-25
Use Query Protocol in storage deal negotiation

### LINKED ISSUE #82: Storage Market should not block reading protocol responses
As either a provider or a client in the storage market, I would like to make reading responses from the other party asynchronous, so that my node does not block waiting on the other party

---

In:
- All Read operations in storage market should be handled in protocol handlers
- No more than one read per handler (i.e should read only incoming value from libp2p)
- After processing read, generate event to state machine
- State handlers take it from there

---

Acceptance criteria:
- Create two retrieval deals with different providers on client. Have one provider stop responding. Other deal should proceed

ingar:
## Problem
Holding stream connections open is not sustainable, and doesn't help with resuming interrupted deals.

## Solution
Use the query protocol in the Client/Provider FSMs, close the deal stream once we get a response.

Resolves #82


storagemarket/client.go
```diff
@@ -38,7 +38,7 @@ type StorageClient interface {
 	GetAsk(ctx context.Context, info StorageProviderInfo) (*SignedStorageAsk, error)
 
 	// GetProviderDealState queries a provider for the current state of a client's deal
-	GetProviderDealState(ctx context.Context, info StorageProviderInfo, proposalCid cid.Cid) (*ProviderDealState, error)
+	GetProviderDealState(ctx context.Context, deal ClientDeal) (*ProviderDealState, error)
```
hannahhoward, line 41: I feel pretty strongly this api should be:
```
GetProviderDealState(ctx context.Context, proposalCid cid.Cid) (*ProviderDealState, error)
```
you can make a private method on Provider with the current signature so you can use it for the state machine.
and then the implementation of GetProviderDealState should be pretty trivial:
```golang
func (c *Client) GetProviderDealState(ctx context.Context, proposalCid cid.Cid) (*storagemarket.ProviderDealState, error) {
var deal ClientDeal
err := c.statemachines.Get(proposalCID).Get(&deal)
if err != nil {
return nil, err
}
return c.getProviderDealState(ctx, deal)
}

File: storagemarket/README.md [hunk 2]
```diff
@@ -9,7 +9,7 @@ data storage.
 * [Installation](#Installation)
 * [Operation](#Operation)
 * [Implementation](#Implementation)
-    * [StorageFunds](#StorageFunds)
+    * [StorageCommon](#StorageCommon)
```
hannahhoward, line 12: Thank you for fixing this!

hannahhoward: +9,999,999

storagemarket/impl/providerstates/provider_states.go
```diff
@@ -229,22 +229,7 @@ func WaitForPublish(ctx fsm.Context, environment ProviderDealEnvironment, deal s
...
-			State:          storagemarket.StorageDealProposalAccepted,
-			Proposal:       deal.ProposalCid,
-			PublishMessage: deal.PublishCid,
-		})
-
-		if err != nil {
-			return ctx.Trigger(storagemarket.ProviderEventSendResponseFailed, err)
-		}
-
-		if err := environment.Disconnect(deal.ProposalCid); err != nil {
```
hannahhoward, line 242: seems like disconnect code should move to RunDealDecider?
Also I believe you should take a look at ProviderEventDealPublished -- I believe that in that code it sets connectionClosed and you should probably move that to whatever the success Event on RunDealDecider.

storagemarket/impl/clientstates/client_states.go
```diff
@@ -136,35 +144,44 @@ func WaitingForDataRequest(ctx fsm.Context, environment ClientDealEnvironment, d
...
 
-// VerifyDealResponse reads and verifies the response from the provider to the proposed deal
-func VerifyDealResponse(ctx fsm.Context, environment ClientDealEnvironment, deal storagemarket.ClientDeal) error {
-	resp, err := environment.ReadDealResponse(deal.ProposalCid)
+// CheckForDealAcceptance is run until the deal is sealed and published by the provider, or errors
+func CheckForDealAcceptance(ctx fsm.Context, environment ClientDealEnvironment, deal storagemarket.ClientDeal) error {
+	dealState, err := environment.GetProviderDealState(ctx.Context(), deal)
 	if err != nil {
-		return ctx.Trigger(storagemarket.ClientEventReadResponseFailed, err)
+		log.Warnf("error when querying provider deal state: %w", err) // TODO: at what point do we fail the deal?
```
hannahhoward, line 151: I'm thinking the proper thing here would be to have a maximum fail count.

storagemarket/integration_test.go
```diff
@@ -221,11 +227,14 @@ func TestMakeDealNonBlocking(t *testing.T) {
 
 	result := h.ProposeStorageDeal(t, &storagemarket.DataRef{TransferType: storagemarket.TTGraphsync, Root: h.PayloadCid})
 
-	time.Sleep(time.Millisecond * 500)
+	wg := sync.WaitGroup{}
```
hannahhoward, line 230: Yay for dropping time.Sleep!

storagemarket/impl/clientstates/client_states.go
hannahhoward: here's a suggestion: 
make waitAgain take an error value -> pass on to ClientEventWaitForDealState
in the FSM code, for if ClientEventWaitForDealState, if err != nil, also increment another value -- say "PollErrorCount".

in this code, add a check for deal.PollErrorCount > some constant (maybe 10? -- miner offline for 5 minutes seems bad) and if so, instead of waiting again, trigger ClientEventReadResponseFailed like before and add back in CheckForAcceptance as an acceptable transition state.

storagemarket/impl/clientstates/client_states.go
```diff
@@ -98,6 +101,11 @@ func WaitingForDataRequest(ctx fsm.Context, environment ClientDealEnvironment, d
 		return ctx.Trigger(storagemarket.ClientEventReadResponseFailed, err)
 	}
 
+	err = environment.CloseStream(deal.ProposalCid)
+	if err != nil {
```
hannahhoward, line 105: Make sure connection closed is being set properly based on this

storagemarket/impl/clientstates/client_states.go
ingar: I wonder about this; the Miner can be totally unresponsive to queries, but can still fulfill their side of the storage deal.  In that case, if we fail the deal on the Client but the Miner finishes sealing and publishing...

storagemarket/impl/providerstates/provider_states.go
```diff
@@ -356,22 +336,6 @@ func FailDeal(ctx fsm.Context, environment ProviderDealEnvironment, deal storage
 
 	log.Warnf("deal %s failed: %s", deal.ProposalCid, deal.Message)
 
-	if !deal.ConnectionClosed {
```
hannahhoward, line 359: So: blocking issue here -- 

This code is used to send a response when the deal is rejected for whatever reason. I believe you still need connection closed for the moment but you need to move it the FSM state change for ProviderEventDataRequested.

And then leave this code here for now

hannahhoward: The more radical solution I proposed is take the deal validation & decision code out of the statemachine entirely.

hannahhoward: alternatively, on more possible solution is to add a new state StorageDealSendRejectionResponse and put this code inside of it before moving on to StorageDealFailing. This would enable you to still delete connection closed.

storagemarket/testnodes/testnodes.go
```diff
@@ -196,6 +198,10 @@ type FakeClientNode struct {
 	OnDealSlashedEpoch         abi.ChainEpoch
 }
 
+func (n *FakeClientNode) SignBytes(ctx context.Context, signer address.Address, b []byte) (*crypto.Signature, error) {
```
hannahhoward, line 201: Can't you just use the implemntation in FakeCommonNode?

PR #305, opened 2020-06-30
Fast retrieval support

### LINKED ISSUE #285: StorageClient can request a fast retrieval
As a StorageClient, I would like to specify that I'm requesting my storage deal be retrievable quickly, so a miner knows to maintain unsealed data

Implementation:
- This is completely off-chain - no changes to spec-actors types
- Augment ClientDeal & MinerDeal with attribute to specify fast retrieval
- Augment Proposal type in **network** module to also have fast retrieval boolean
- Augment StorageProviderNode's OnDealComplete to take fast retrieval boolean

Out of scope:
- Any decision logic around fast retrieval -- assume for now anything that changes based on fast retrieval is handled by custom deal decider
- Lotus integration (need to verify if storage-fsm is setup to use this first

ingar:
## Summary
Send the fast retrieval flag along with the proposal and track it in the `ClientDeal` and `MinerDeal`.  Also return this flag in the deal status query.

Resolves #285

### DIFF
--- a/storagemarket/impl/client.go
+++ b/storagemarket/impl/client.go
@@ -346,6 +346,7 @@ func (c *Client) ProposeStorageDeal(ctx context.Context, addr address.Address, i
 		Miner:              info.PeerID,
 		MinerWorker:        info.Worker,
 		DataRef:            data,
+		FastRetrieval:      fastRetrieval,
 	}
 
 	err = c.statemachines.Begin(proposalNd.Cid(), deal)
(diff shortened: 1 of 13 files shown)


hannahhoward:
Otherwise LGTM

ingar:
Note about changing the `StorageProviderNode` interface.  That method gets passed a `MinerDeal`, which now has a `FastRetrieval` bool in it.  I'm passing the FastRetrieval flag in via this existing parameter; it will avoid interface churn, unless we need the interface change for some other reason.

PR #307, opened 2020-07-01
Validate client's DataCap for verified deals

### LINKED ISSUE #286: Provider Rejects Verified Deals At Proposal Time
As a provider, I want to reject a storage deal at the time it's proposed if VerifiedDeal is requested but the client is not verified or has exceeded it's data cap.

Steps:
- Add method to StorageProviderNode - `GetDataCap(ctx context.Context, address.Address, tok shared.TipSetKey) (verifreg.DataCap, error)
- When validating a deal proposal in the provider:
   - if VerifiedDeal = true on deal proposal
      - Call GetDataCap method with client address
      - if err != nil, reject deal (most likely client not verified)
      - if DataCap < deal piecesize, reject deal

ingar:
## Summary
For verified deal proposals, reject the deal if the `verifreg.DataCap` for the client isn't large enough.

Resolves #286


hannahhoward:
This looks done-ish -- is it done?

PR #309, opened 2020-07-02
Run deal restarts in go routine

hannahhoward:
# Goals

Deal restarts make take a while particularly if there are connection timeouts. There's no reason to
run them synchronously in node initialization, and we definitely should not fail node startup if there are errors.

# Implementation

Run restart deals in a go routine, simply log any errors


hannahhoward:
@arajasek I think I'm gonna create a meta issue in markets that's like "deal reliability tracking" cause there are several things we may want to do. Note the current version of markets does not attempt to reopen connections btw cause the newest protocol has no long running libp2p connections for deals.

PR #316, opened 2020-07-06
Protect/tag libp2p connections during data transfer

ingar:
## Problem
Connection/peer tagging was removed in a previous commit, but it is needed to prevent connections from being recycled.

## Summary
Add peer tagging during data transfer in both the client and provider FSMs.


ingar:
Closing in favor of https://github.com/filecoin-project/go-data-transfer/pull/55

PR #323, opened 2020-07-15
(WIP - don't review yet) Store PayloadCID in flexible DealProposal.Label field

mgoelzer:
This PR adds the Payload CID, a hash computed over the client's data in its original form, into the flexible `Label` field of each storage deal's `DealProposal` struct.  This ensures that each storage deal's Payload CID will be published on the blockchain together with that deal's corresponding Piece CID.

By publishing messages containing both CIDs, the blockchain will become a record of {PayloadCID, PieceCID} tuples that have existed on the Filecoin network.  This will enable a variety of indexing and retrieval-related use cases, such as: 
- retrieval by 3rd parties, such as retrieval market participants, even when they have only partial information about the original storage deal, 
- retrieval by clients who no longer possess all of their storage deal's original metadata,
- the generation of blockchain indices that keep track of alternate names that a piece of data on the Filecoin network might be known as (e.g., to support Filecoin network light clients),
- other proxy retrieval scenarios.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/filecoin-project/go-statestore v0.1.0
 	github.com/filecoin-project/go-storedcounter v0.0.0-20200421200003-1c99c62e8a5b
 	github.com/filecoin-project/sector-storage v0.0.0-20200615154852-728a47ab99d6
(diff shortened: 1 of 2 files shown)


hannahhoward:
closed as this change was merged elsewhere

PR #332, opened 2020-07-23
Adding different prices for verified deals

mrsmkl:
This is one way to implement this, adding it here so that it won't be forgotten.
Probably there should be some more generic way to implement pricing.


arajasek:
Thanks so much for this! I'm going to land this with minor changes in #347

PR #333, opened 2020-07-23
Upgrade specs-actors

hannahhoward:
# Goals

Match spec actors sha in Lotus/Next

# Implementation

- upgrade deps
- fix some cbor-gen issues
- fix checks of nil selectors on retrieval proposals
- change local discovery back to using cborutil -- it's super error prone vis a vis running CBOR-gen otherwise


retrievalmarket/types.go
```diff
@@ -237,6 +237,10 @@ type Params struct {
 	UnsealPrice             abi.TokenAmount
 }
 
+func (p Params) SelectorSpecified() bool {
+	return p.Selector != nil && !bytes.Equal(p.Selector.Raw, cbg.CborNull)
```
ingar, line 241: Ah nice.

PR #336, opened 2020-07-24
Track deals funding for deals that are being negotiated

### LINKED ISSUE #331: Track total funds commited in SMA, Client and Provider
As a client (using StorageClient software) or a miner (StorageProvider) I would like to make sure that I have enough funds available in the market actor for all of my current known deals, so that if I make two simultaneous deals, they both succeed.

Background:
When money is put into the SMA, it is considered "Available"
When a deal is published, funds needed for the deal are locked, and are no longer available.
Currently we use EnsureFunds, which calls MarketEnsureAvailable, which insures the amount past to it is available in the SMA.
Currently, we only submit the value of the current deal to EnsureFunds, but actually, for all deals to go through, we need to have available funds for every deal that is in a state between New and Published (and not failed)

StorageClient & StorageProvider should maintain, in a datastore, a seperate value for total funds committed for our own deals.

Recommend writing a class like StoredAsk -- maybe FundsCommittedTotal
- make sure it has a mutex lock (will be accessed by multiple deals)
- propobly maintain cached value like StoredAsk to avoid datastore lookup on Get
- Should have AddFundsCommitted, ReleaseFundsCommitted, Get (or something like that)

Client:

- In StorageClient ensure funds states, add committed funds for current deal and then call node.EnsureFunds with total commited funds
- Release committed funds probably right after VerifyDealPublished (i.e. at the beginning of VerifyDealActivated) and in any error state

Provider:

On Provider, repeat the same for the provider ensure funds steps and release funds in StorageDealStaged & error states (at the beginning of handoff deal probably)

Notes:

- Should probably not require changes to statemachine for the most part
- Should not require changes to node interface

ingar:
## Summary
Add a new interface, `DealFunds`, and use it to track outstanding funds needed for storage deals that are in progress.

Resolves #331


storagemarket/impl/clientstates/client_states.go
```diff
@@ -46,9 +48,19 @@ func EnsureClientFunds(ctx fsm.Context, environment ClientDealEnvironment, deal
...
-	mcid, err := node.EnsureFunds(ctx.Context(), deal.Proposal.Client, deal.Proposal.Client, deal.Proposal.ClientBalanceRequirement(), tok)
+	requiredFunds, err := environment.DealFunds().Reserve(deal.Proposal.ClientBalanceRequirement())
+	if err != nil {
+		return ctx.Trigger(storagemarket.ClientEventEnsureFundsFailed, xerrors.Errorf("tracking deal funds: %w", err))
+	}
+
+	mcid, err := node.EnsureFunds(ctx.Context(), deal.Proposal.Client, deal.Proposal.Client, requiredFunds, tok)
 
 	if err != nil {
+		_, err2 := environment.DealFunds().Release(deal.Proposal.ClientBalanceRequirement())
```
hannahhoward, line 59: This needs to happen in the global fail state handler (FailDeal) - any error that happens between when EnsureFunds is called and the deal is published need to release the funds

hannahhoward: to know whether to release in FailDeal may mean recording a local value of funds reserved  on the deal struct -- starts at 0 when deal is created, becomes ClientBalanceRequirement here, goes back to 0 after ValidateDealPublished, and then in the FailDeal handler, if reserved > 0, release those funds

storagemarket/impl/providerstates/provider_states.go
```diff
@@ -170,9 +172,20 @@ func EnsureProviderFunds(ctx fsm.Context, environment ProviderDealEnvironment, d
...
-	mcid, err := node.EnsureFunds(ctx.Context(), deal.Proposal.Provider, waddr, deal.Proposal.ProviderCollateral, tok)
+	requiredFunds, err := environment.DealFunds().Reserve(deal.Proposal.ProviderCollateral)
+	if err != nil {
+		return ctx.Trigger(storagemarket.ProviderEventTrackFundsFailed, xerrors.Errorf("reserving deal funds: %w", err))
+	}
+
+	mcid, err := node.EnsureFunds(ctx.Context(), deal.Proposal.Provider, waddr, requiredFunds, tok)
 
 	if err != nil {
+		_, err2 := environment.DealFunds().Release(deal.Proposal.ProviderCollateral)
```
hannahhoward, line 183: again this needs to happen in the global fail state handler (`FailDeal`) -- same comment as client

PR #356, opened 2020-08-05
Properly add multiaddrs to avoid dialing issues

hannahhoward:
# Goals

Handle MultiAddrs properly in storage market, retrieval market to avoid dialing issues

# Implementation

- Restore calling GetMinerInfo on the storage side when proposing and resuming deals
- Up the timeout on adding multiaddrs
- Add peerstore adding functionality to retrieval market network
- Add a function on the RetrievalClientNode to get known addresses for a retrieval peer
- Convert Retrieve function to take a RetrievalPeer
- Add multiaddrs on both RetrievalQuery and Retrieve


hannahhoward:
@ingar the assumption is with 8 HR TTL that should be enough to complete the current deal and every new deal/queery will readd the values.

PR #369, opened 2020-08-13
Include correction factor in DealProviderCollateralBounds

arajasek:
This will help, but I think the question of "how to estimate DealProviderCollateralBounds in the future" needs to be answered.

We also might wanna make the `correctPercentage` easily configurable (I can see miners needing to quickly increase that % under certain network conditions).


hannahhoward:
also let's definitely put these numbers as constants not hard coded into the code, and it would be nice to allow them to be configurable (through StorageClientOption / StorageProviderOption) so if our initial numbers don't work out super well miners can customize them to their needs.

arajasek:
Closing this in favour of making the changes directly in Lotus's adapters. We might wanna reopen this discussion once we have a better sense of how we should be estimating provider collateral bounds (how we should be expecting the number to change between the client creating the proposal and the provider publishing the deal on chain).

PR #374, opened 2020-08-17
switch over to map encodings for all markets structs

whyrusleeping:
Since we're already making one format breaking change, i figured we could make another one, to prevent the previous breaking change from being a breaking change in the future. If we use map encodings (like we do in the storage-fsm) then adding a new field doesnt break everything else.


whyrusleeping:
This also allows us to implement data migrations more easily in the future, we could add version fields to structs we need to upgrade, and handle version migration in code

whyrusleeping:
cbor-gen-check wants this PR to be merged first: https://github.com/hannahhoward/cbor-gen-for/pull/1

hannahhoward:
@whyrusleeping merged other PR

hannahhoward:
hey @whyrusleeping what do you want to do about this now that Space Race is out there?

hannahhoward:
This was shipped in the migrations release

PR #375, opened 2020-08-20
Clients must error when provider errors

hannahhoward:
# Goals

Due to both a bug in go-fil-markets and a seperate bug in data transfer (fixed here: https://github.com/filecoin-project/go-data-transfer/pull/67), when a provider errored during retrieval, the client would simply hang, causing a cascading set of failures.

# Implementation

- Only major code fix is a missing event in the client FSM
- Update data transfer with fix in PR
- Add test to verify


hannahhoward:
no we don't. :( not yet anyway. need to figure that out for data transfer.

PR #381, opened 2020-08-27
Fix restart issue with deal transferring

### LINKED ISSUE #378: Deal restarts fails if Storage Client crashes in the StorageDealTransferring state
**Problem**

* Client calls the `InitiateDataTransfer` method which triggers an `ClientEventDataTransferInitiated` event. 
* This event moves the client to the `StorageDealTransferring` state and the provider to the `StorageDealWaitingForData` state if the client is able ti successfully open a Push channel to the client.
* When the client and provider restart, neither can make any progress because neither the `StorageDealTransferring` state, nor the `StorageDealWaitingForData` state have any handlers on the client and provider respectively.

**Solution**
Upon restart, move all deals on the client that are in the `StorageDealTransferring` state to the `StorageDealStartDataTransfer` state to trigger a data transfer.

aarshkshah1992:
Closes #378 .

### DIFF
--- a/docs/storageclient.mmd
+++ b/docs/storageclient.mmd
@@ -57,6 +57,7 @@ stateDiagram-v2
 	7 --> 8 : ClientEventDealExpired
 	7 --> 26 : ClientEventDealCompletionFailed
 	11 --> 26 : ClientEventFailed
+	17 --> 16 : ClientEventRestart
 
 	note left of 3 : The following events only record in this state.<br><br>ClientEventFundsReleased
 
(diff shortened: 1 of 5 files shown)


aarshkshah1992:
@hannahhoward  Let me know if we need to test the restarts after more events.

hannahhoward:
@aarshkshah1992 I would check ClientEventDealPublished and ClientEventDealActivated as well, though you can do so in a seperate ticket. That will also probably require making sure the node methods to not return immediately, as I imagine the final states are moved through in quick sucession during test (in a real environment, it's not quick)

hannahhoward (approved):
Please run `make prepare-pr` before merging to pass the docs-check

hannahhoward:
@aarshkshah1992 also TestMakeDealNonBlocking being flaky is a known issue -- if you want to look at that in a seperate ticket you're welcome to.

aarshkshah1992:
Created issue to track the remaining tests:

https://github.com/filecoin-project/lotus/issues/3443

PR #389, opened 2020-09-02
Persist cids recieved from the data transfer channel

aarshkshah1992:
For filecoin-project/lotus#3417.

@hannahhoward This is step 1 which persists and cleans-up the received cids in the provider.

This PR is against a long running feature branch.

Depends on https://github.com/filecoin-project/go-data-transfer/pull/71.

### DIFF
--- a/storagemarket/impl/dtutils/dtutils.go
+++ b/storagemarket/impl/dtutils/dtutils.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 
 	"github.com/ipfs/go-cid"
+	"github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-graphsync/cidset"
 	logging "github.com/ipfs/go-log/v2"
 	"github.com/ipld/go-ipld-prime"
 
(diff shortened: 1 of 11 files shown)


storagemarket/impl/dtutils/dtutils.go
```diff
@@ -56,11 +61,37 @@ func ProviderDataTransferSubscriber(deals EventReceiver) datatransfer.Subscriber
 			if err != nil {
 				log.Errorf("processing dt event: %w", err)
 			}
+		case datatransfer.DataReceived:
+			// TODO Should we memoize here ? This can get inefficient as this will event will keep firing for the same cid if we recieve the same cid
+			// multiple times
+			cids := channelState.ReceivedCids()
+			if len(cids) != 0 {
+				set := cid.NewSet()
```
hannahhoward, line 69: This strikes me as potentially ineffecient -- to reconstruct the CID set on every block? At minimum it's an n^2 operation.

Data transfer state is persisted to disk. I would strongly consider, given that you are already recording received CIDs there, that you avoid recording them here as well.

It's your call, but think you are going to be happier if you simply build an API on data transfer's manager called:

```
	RestartDataTransferChannel(ctx context.Context, chid ChannelID) error
```

Anyway, I'm fine if you wanna proceed -- your call, I just think you will find yourself getting a bit stuck down this path.

If you'd like to keep the recording in here, I'd recommend taht we not record the whole list of CIDs on the data transfer -- and just provide a "LastCID" or something that updates on every DataReceived.

hannahhoward (approved):
See commends about how we record data

aarshkshah1992:
This is the wrong appraoch.

PR #392, opened 2020-09-04
Non blocking retrieval

hannahhoward:
# Goals

Make retrieval faster by only blocking for chain operations when:
1. a channel does not exist
2. we run out of funds in the channel to make vouchers

Also, fill in three functions on the retrieval client which were just implemented as "panic" calls which seems like poor form for code going into feature freeze soon.

# Implementation

Previously we had two scenarios:
1. Start a retrieval between A & B -> no payment channel exists. We would then create a payment channel with funds in it, and wait for that to finish, then proceed
2. Start a retrieval between A&B -> Payment channel exists. We would then add funds to the payment channel, wait for that to finish, then proceed

The main chain has to do with scenario 2. Now, our steps are:
- Add funds to the payment channel, but don't wait for that to finish
- Start the retrieval
- When we attempt to create the voucher, and find we don't have sufficient funds (if that happens at all), now wait for the finish

However, this actually also affects scenario 1, because we might actually deplete the funds in the channel we create if we're running a simultaneous retrieval on the same channel where we simply added funds but didn't wait. Because both retrievals are draining the channel, it's possible the retrieval that created the channel might hit a point where a voucher can't create first.

So now our algorithm for both scenarios is:
- create a channel if not present, wait for finish
*or*
- add funds if channel is present, do not wait
- proceed with retrieval until we try to create a voucher and get a shortfall of funds
- now, if we already no about a pending add operation, wait for that
- if we don't have a pending operation, query the channel for
   - balances
   - pending operations
- determine if between current and pending operations will be enough to cover next payment
- if not - go into an insufficient funds state
- if yes, and we have enough to cover immediately, go back to processing retrieval (this won't happen the first time but we'll return to this check if someone adds funds to the channel manually)
- if yes, but we need some of the pending balance, wait for the pending operations then go back to processing retrieval

This returns us to one other missing piece of functionality -- what to do generally if we run out of funds in the channel. While this was not likely while retrieval is the only one using payment channels, as soon as we add potential other users this assumption falls down. We had a previously unimplemented AddFunds() call, which has now been implemented as TryRestartInsufficientFunds -- rather than add funds directly, which seems like it's the payment channel manager's job, we have a call that can be used to attempt to restart deals on a given channel after funds are added.

This PR also implements a few extra peices around one other unimplemented function -- the ability to cancel retrievals.


dirkmc:
Looks great 👍 

Looking at this code I realized that strange things could happen if there are queued funds but not a pending add funds msg cid. So I will change the payment channel code such that CheckAvailableFunds will only ever return a positive queued funds values if there is a pending msg CID (otherwise it will block for a moment until the queued funds become pending).

dirkmc (approved):

ingar (approved):
Looks good.  

This is similar to multiple concurrent deals on the storage side of things.. like retrieval, we aren't tracking total outstanding funds, just making sure that we have enough funds for the current deal, and if something happens in the meantime, we just add funds again.

I have a worry that if LOTS of simultaneous deals are happening (dozens or more), the behavior will be pretty unpredictable (add funds, another deal uses some random part of the funds, over and over).  Anyways, just food for thought.

PR #393, opened 2020-09-04
Convert label JSON to CBOR

hannahhoward:
# Goals

Shorten the label field by using CBOR

# Implementation

Use CBOR instead of json (previously was using different options with JSON encoder to avoid whitespace)


mikeal:
that’s a codec bug, shouldn’t even be allowed to do that.

hannahhoward:
@mikeal you should check in with @warpfork about that -- the default DAGJSON encoder in go-ipld-prime does whitespace

hannahhoward:
anyway, I make this into a CBOR converter instead

PR #398, opened 2020-09-09
Encode cid directly in label field rather than try to use structured data

hannahhoward:
# Goals

Because of issues with encoding the CID in a CBOR struct inside a string field for label, but also not wanting to pad things out with JSON, we're switching to encoding the CID directly as Multibase string.

# Implementation

- Write label field as a multibase string 
   - Use base58 for Cid V0 (only one supported)
   - Use base64 for Cid V1 (base32 is default but we're trying to keep things more compact)


ingar (approved):
Looks good

PR #399, opened 2020-09-10
Test Restart for ClientEventDealPublished AND ClientEventDealActivated

aarshkshah1992:
Closes https://github.com/filecoin-project/lotus/issues/3443.

Also, downgrades stream reset log in `HandleDealStatusStream` to `WARN` instead of `ERROR`.


hannahhoward (changes requested):
Awesome and generally LGTM.

However, I really prefer to not put time.Sleeps' in test data. what I would rather do is have the DelayFakeCommonNode have channels, and if the boolean is true, it reads from the channel. Then, whenever you want to get it unstuck, you write to the channel. (Can just be a chan struct{})

aarshkshah1992:
@hannahhoward Have addressed your review.

aarshkshah1992:
@hannahhoward Have addressed your comments.

hannahhoward (approved):

PR #400, opened 2020-09-10
Store final/actual cid for publish message

ingar:
## Problem
The cid of a message can change when published to the chain because of gas fee changes.  Since the Provider passes a cid back to the client when queried for deal status, it could send the incorrect cid back to the client, causing the client to fail to verify the published deal.

## Solution
Store the actual/final cid in the `MinerDeal`

### Notes
This is the market-side fix for https://github.com/filecoin-project/lotus/issues/3703

### DIFF
--- a/storagemarket/impl/client.go
+++ b/storagemarket/impl/client.go
@@ -433,7 +433,7 @@ func (c *Client) AddPaymentEscrow(ctx context.Context, addr address.Address, amo
 		return err
 	}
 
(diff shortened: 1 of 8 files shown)


hannahhoward (approved):
We should probably do a seperate follow up PR after we get this code into LOTUS where we save the final cids for other calls to wait for publish, just so we have the right information.

But I'm LGTM'ing an merging this for now.

PR #402, opened 2020-09-14
Increase retry timeout

aarshkshah1992:
Increase retry timeout for connection attempts


hannahhoward (approved):
LGTM by me

We are probably gonna have some tests run very long now so we'll need to figure that out eventually

aarshkshah1992:
@hannahhoward 

Yup, looks like `TestProviderStop` is already timing out. Can we go ahead and disable it ?

PR #403, opened 2020-09-17
Configurable Retry Parameters (to speedup tests)

hannahhoward:
# Goals

- Remove need to for further PRs to change network retry parameters by making them a configuration option
- Make tests run quickly again

# Implementation

- add the var args config option pattern to the Libp2p impls for storage and retrieval market networks
- add an options to change all the retry parameters
- set parameters to avoid retries in relevant tests


aarshkshah1992 (approved):
LGTM !

PR #408, opened 2020-09-21
Consider ClientCollateral when validating deal proposal

ingar:
## Problem
Previously, we weren't including the client collateral when considering storage deal proposals; it was assumed it was always 0.

## Solution
Use `ClientBalanceRequirement()` when verifying client funds, which includes the client collateral.

Resolves filecoin-project/lotus/issues/3864


hannahhoward (approved):
LGTM

PR #410, opened 2020-09-21
Persist retrieval ask on disk

ingar:
## Summary
Create a retrievalmarket `AskStore` interface and implementation, which will persist the retrieval ask to disk so that the configuration will persist through restarts.  This is modeled after the storagemarket's `StoredAsk`.

Resolves filecoin-project/lotus/issues/3907


hannahhoward (approved):
LGTM

PR #411, opened 2020-09-23
Resume StorageMarket Provider Data Transfer

aarshkshah1992:
For https://github.com/filecoin-project/lotus/issues/3417.


storagemarket/integration_test.go
```diff
@@ -286,6 +286,97 @@ func TestMakeDealNonBlocking(t *testing.T) {
...
+
+	expireWg.Add(1)
+	_ = client.SubscribeToEvents(func(event storagemarket.ClientEvent, deal storagemarket.ClientDeal) {
+		if event == storagemarket.ClientEventDealExpired {
+			expireWg.Done()
+		}
+	})
+
+	// sleep so go-data-transfer gives up on retries after creating new connection
+	time.Sleep(15 * time.Second)
```
hannahhoward, line 356: is this waiting for go-graphsync or go-data-transfer to give up? I think go-graphsync, right?

that's fine. I should provide a way to disable retries in graphsync to prevent this.

anytime we're putting in a sleep longer than like a second in a test we may want put a check to skip in `--short` mode so you can run all tests locally with some expectation they'll run relatively fast.

```
   if testing.Short() {
        t.Skip("skipping test in short mode.")
    }
```

storagemarket/impl/providerstates/provider_states_test.go
```diff
@@ -401,6 +403,57 @@ func TestEnsureProviderFunds(t *testing.T) {
...
+		"RestartDataTransfer errors": {
+			dealParams: dealParams{
+				TransferChannelId: &channelId,
+			},
+			environmentParams: environmentParams{
+				RestartDataTransferError: xerrors.New("some error"),
+			},
+			dealInspector: func(t *testing.T, deal storagemarket.MinerDeal, env *fakeEnvironment) {
+				require.Eventually(t, func() bool {
+					fmt.Printf("\n deal state is %s", storagemarket.DealStates[deal.State])
```
hannahhoward, line 442: This isn't going to work unfortunately.

The problem is really hard to see but the state of the deal only updates when you call ReplayEvents on the FSM TestContext (see line 1228 of this file)

Sorry I hadn't thought about any of these tests being asynchronous. 

Fixing this is non-trivial. Given that, I'm ok if you comment out the `RestartDataTransfer errors` test for now (leave the code there till we can figure something out)

hannahhoward: You'll see if you look at the CI that this test is failing -- my guess is if it works on your local machine it's just cause your local machine is faster

hannahhoward (changes requested):

hannahhoward (changes requested):
Sorry please comment out the `RestartDataTransfer errors` test for now

aarshkshah1992:
@hannahhoward Looks good now. Please let me know what you think.

PR #415, opened 2020-09-25
Move PieceStore To Map Encodings

hannahhoward:
# Goals

Switch to map encoding for piecestore, safely without breaking existing data

# Implementation

This is the first of several introductions of data store migrations.

It's important to review https://github.com/filecoin-project/go-ds-versioning (unfortunately, not yet well documented)
go-ds-versioning provides various tools:
- You can define datastore migrations via an individual record transformation function -- i.e. func(oldRecord oldStruct) (newStruct, error)
- You can setup migrations that have an "old version" and a "new version"
- You can setup versioned datastores, statestores, and fsm groups -- these have their normal parameters, plus a list of migrations, and a target version -- go-ds-versioning figure out how best to migrate to the target version based on the current version (versions are namespaces wrapped under a key, and there is code to migrate an unversioned store to a versioned store). When you setup datastores this way, you have to run a migrate function -- until you do, all of the functions provided by the interface return errors (this will allow us to let Lotus startup while leaving the markets in a disabled state until migrations complete)

The changes here to the piece store are as follows:
- Switch to map-encoding for the current piecestore types
- Add a migrations package that holds all the old types (actually, they're identical, just tuple encoded), and define functions to transform CIDInfo and PieceInfo (the top level ones that are the records of the database)
- Define migration lists for both stores
- Move the piecestore implementation to a seperate impl package to resolve import cycles
- Modify the piecestore constructor to setup versioned stores
- Add a Start function to piecestore to run migrations on stores
- Add an OnReady function to listen for when the piecestore is done migrating (used only in tests so far -- probably this will just be a log in Lotus)
- Update tests infrastructure elsewhere


ingar (approved):
Tuple to map encoding and migration. 👍

PR #419, opened 2020-09-28
early return in retrieval validation to prevent unnecessary disk access

dylanlott:
I believe I found a path where a malicious user could make a GetPiece request that is outside of the providing node's pricing limits, but still force a disk check to load up the PieceCID and all following CIDs without having to actually pay for the access. Initially thought that this would just be a good suggestion but after closer inspection I believe this is a potential DoS vector. A single node could send a large number of invalid requests for a piece that the provider node actually does store, and trigger the same number of unnecessary disk accesses.

acceptDeal => GetPiece (which checks that the piece ID is not nil but in the case of it being valid date, it wouldn't be nil) => getPieceInfoFromCid => pieceStore.PieceStore interface => which, as far as I can tell, causes a block access of all the blocks related for that store. The more the merrier, but this could actually be even worse because a malicious user could store a large CID and force access of it (the larger the file, the more underlying block accesses) which makes the attack curve even sharper with file size.

### DIFF
--- a/retrievalmarket/impl/requestvalidation/requestvalidation.go
+++ b/retrievalmarket/impl/requestvalidation/requestvalidation.go
@@ -5,19 +5,17 @@ import (
 	"context"
 	"errors"
 
+	datatransfer "github.com/filecoin-project/go-data-transfer"
+	"github.com/filecoin-project/go-fil-markets/piecestore"
+	"github.com/filecoin-project/go-fil-markets/retrievalmarket"
+	"github.com/filecoin-project/go-fil-markets/shared"
+	"github.com/filecoin-project/go-multistore"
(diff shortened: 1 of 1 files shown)


hannahhoward:
I'm fine with this fix, but I am curious: 

what’s to prevent someone to just submit a ton of requests with reasonable pricing parameters and then hang up as soon as the deal is accepted? we have the unseal price to prevent someone from continually hanging up after making the provider spend the cost to unseal (a much much larger cost), but it seems hard to imagine we can truly mitigate any disk access whatsoever prior to demanding payment. It seems like we may just have to have some kind of protection when we get spammed by the same person lots and lots of times.

hannahhoward:
Closed but merged in rebased PR

PR #420, opened 2020-09-29
Switch to cbor map encoding for storage market

hannahhoward:
# Goals

- Move all storage data structures to CBOR Map encoding
- Setup transparent migrations (continue to implement ds-versioning)
- Don't break network protocols -- continue to support old versions (for now) -- as an exercise in proving how to do non breaking upgrades

# Implementation

- First setup all migrations, similar to style of migrations in piecestore
- Setup and test client and provider for storage to work with versioned datastores
- Also migrate the StoredAsk data structure
- Move around the test harness so the code is shared
- Setup all protocols (ask, deals, deal status) to accept multiple stream options for both versions of protocol
- We have to change some signatures for the network streams cause they have signatures -- unfortunately, the signature on a tuple struct is different than a map encoded one... so the signature bytes actually change. We pass in a resigning function for write functions (we can re-sign cause we know our own key), and we return the original bytes for read functions (we cannot resign, cause it's the other party's data structure -- be can vouch for the fact that the original bytes, if they match the signature, do match the return unsigned map-encoded structure in terms of parameters
- Add options to disable various networking options so we can test and verify our changes work


ingar (approved):
Migrate FTW.

PR #423, opened 2020-10-02
Handle re-sent deal proposals

ingar:
## Problem
There is an edge case in the deal flow where a client sends the deal proposal and shuts down for some reason.  When the client restarts, it resends the deal proposal, but the provider has already started tracking the deal, and rejects the proposal.

## Solution
There is no problem with re-sending the deal response, and not changing anything in the provider state.  This will allow the deal to continue on the client's state machine.

Resolves part of https://github.com/filecoin-project/lotus/issues/3966


hannahhoward (approved):
LGTM

PR #424, opened 2020-10-02
Fix flaky TestRestartClient/ClientEventDataTransferComplete

ingar:
## Summary
Fix this particular flaky test by providing a specific event to stop the provider at, rather than stopping the provider when the client event is hit.

## Note
There is still a slim edge case here, if the provider shuts down after a FSM event is emitted and before the state transition happens.  However, this edge case, while rare in actual usage, is one that affects all events and will have to be addressed in the FSM.

See https://github.com/filecoin-project/lotus/issues/3966


hannahhoward (approved):
LGTM

PR #427, opened 2020-10-08
Providers should reject deals with invalid durations

arajasek:
Addresses https://github.com/filecoin-project/lotus/issues/3858


arajasek:
Can't seem to get the imports check to pass (/scripts/fiximports seems to edit very file there is)

PR #429, opened 2020-10-09
openStream: avoid time.Sleep in exponential backoff

jsign:
The backoff logic was doing `time.Sleep(...)` which isn't nice for handling cancelation.

### DIFF
--- a/retrievalmarket/network/libp2p_impl.go
+++ b/retrievalmarket/network/libp2p_impl.go
@@ -107,8 +107,11 @@ func (impl *libp2pRetrievalMarketNetwork) openStream(ctx context.Context, id pee
 		if nAttempts == impl.maxStreamOpenAttempts {
 			return nil, xerrors.Errorf("exhausted %d attempts but failed to open stream, err: %w", int(impl.maxStreamOpenAttempts), err)
 		}
-		d := b.Duration()
-		time.Sleep(d)
+		select {
+		case <-ctx.Done():
+			return nil, xerrors.Errorf("backoff canceled by context")
+		case <-time.After(b.Duration()):
+		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


retrievalmarket/network/libp2p_impl.go
```diff
@@ -107,8 +107,11 @@ func (impl *libp2pRetrievalMarketNetwork) openStream(ctx context.Context, id pee
 		if nAttempts == impl.maxStreamOpenAttempts {
 			return nil, xerrors.Errorf("exhausted %d attempts but failed to open stream, err: %w", int(impl.maxStreamOpenAttempts), err)
 		}
-		d := b.Duration()
-		time.Sleep(d)
+		select {
+		case <-ctx.Done():
+			return nil, xerrors.Errorf("backoff canceled by context")
+		case <-time.After(b.Duration()):
```
Stebalien, line 113: Ideally use `time.NewTimer` to avoid (temporarily) [leaking](https://golang.org/pkg/time/#After) a timer on cancellation. But this isn't a huge deal.

Stebalien (changes requested):

Stebalien (approved):

retrievalmarket/network/libp2p_impl.go
jsign: Yeah, considering that but wouldn't imagine would be a big issue having the timers hanging around a bit. 
I'll change to cancel them.

Stebalien (approved):

Stebalien:
I mostly just pedantically avoid `time.After` except when testing, so I don't have to think about this when reading code.

hannahhoward:
Merged in https://github.com/filecoin-project/go-fil-markets/pull/431

PR #430, opened 2020-10-12
Resume Storage Market Data Transfer

aarshkshah1992:
PR to merge data transfer work to master.


storagemarket/testharness/dependencies/dependencies.go
```diff
@@ -108,7 +109,8 @@ func NewDependenciesWithTestData(t *testing.T, ctx context.Context, td *shared_t
 	dtTransport1 := dtgstransport.NewTransport(td.Host1.ID(), gs1)
 	dt1, err := dtimpl.NewDataTransfer(td.DTStore1, td.DTNet1, dtTransport1, td.DTStoredCounter1)
 	require.NoError(t, err)
-	err = dt1.Start(ctx)
+	testutil.StartAndWaitForReady(ctx, t, dt1)
```
hannahhoward, line 112: Hey you don't actually need to use this version -- the function is already present in the shared_testutil for go-fil-markets and will work with go-data-transfer. I should probably extract that whole functionality to common utilities repo at some point :)

storagemarket/dealstatus.go
```diff
@@ -65,6 +65,14 @@ const (
 	// StorageDealTransferring means data is being sent from the client to the provider via the data transfer module
 	StorageDealTransferring
 
+	// StorageDealProviderTransferRestart means a storage deal data transfer from client to provider will be restarted
```
hannahhoward, line 68: at this point, we should put these at the end of the states list -- people are starting to rely on the deal states list not changing

hannahhoward (changes requested):
I think you're maybe still working on this, but here are two requests

storagemarket/impl/clientstates/client_fsm.go
```diff
@@ -65,20 +66,39 @@ var ClientEvents = fsm.Events{
...
 	fsm.Event(storagemarket.ClientEventDataTransferInitiated).
-		From(storagemarket.StorageDealStartDataTransfer).To(storagemarket.StorageDealTransferring),
+		FromMany(storagemarket.StorageDealStartDataTransfer).To(storagemarket.StorageDealTransferring).
+		Action(func(deal *storagemarket.ClientDeal, channelId datatransfer.ChannelID) error {
+			deal.TransferChannelID = channelId
+			return nil
+		}),
+
+	fsm.Event(storagemarket.ClientEventDataTransferRestarted).
+		FromMany(storagemarket.StorageDealClientTransferRestart, storagemarket.StorageDealStartDataTransfer).To(storagemarket.StorageDealTransferring).
```
hannahhoward, line 99: This should not ideally run when in StorageDealStartDataTransfer right? Why is this in the list?

storagemarket/impl/providerstates/provider_fsm.go
```diff
@@ -39,7 +40,23 @@ var ProviderEvents = fsm.Events{
...
+
+	fsm.Event(storagemarket.ProviderEventDataTransferRestartFailed).
+		From(storagemarket.StorageDealProviderTransferRestart).To(storagemarket.StorageDealFailing).
+		Action(func(deal *storagemarket.MinerDeal, err error) error {
+			deal.Message = xerrors.Errorf("error restarting data transfer: %w", err).Error()
+			return nil
+		}),
+
+	fsm.Event(storagemarket.ProviderEventDataTransferRestarted).
+		FromMany(storagemarket.StorageDealWaitingForData, storagemarket.StorageDealProviderTransferRestart).To(storagemarket.StorageDealTransferring).
```
hannahhoward, line 57: What's the scenario where this is called in StorageDealWaitingForData?

storagemarket/types.go
```diff
@@ -102,27 +103,30 @@ type MinerDeal struct {
 
 	DealID       abi.DealID
 	CreationTime cbg.CborTime
+
+	TransferChannelId *datatransfer.ChannelID
```
hannahhoward, line 107: On the provider side, we use a pointer to a channelID, while on the client side, we use a channelID value -- what's the rationale for the difference?

hannahhoward (approved):

storagemarket/types.go
aarshkshah1992: I think this is an oversight. Let's use pointer on both sides.

storagemarket/impl/providerstates/provider_fsm.go
aarshkshah1992: @hannahhoward 

If the provider crashes after the channel is open on it's side but before seeing the corresponding event in the `ProviderDataTransferSubscriber`,  it will never enter the `StorageDealTransferring` state.

storagemarket/impl/clientstates/client_fsm.go
aarshkshah1992: @hannahhoward 

We now move from `StorageDealStartDataTransfer ` to `StorageDealTransferring` only after we **see** the `Accept` from the other side.

However, what if the other side sends an accept, we crash before receiving it and then restart. It means the channel exists on both sides and yet, we are restarting in the `StorageDealClientTransferRestart ` state.

PR #432, opened 2020-10-13
Record data transfer errors

hannahhoward:
# Goals

- Provide more information in deals about the state of disconnects

# Implementation

- in storage market, record an error when data transfer stalls
- in retrieval market, cause we have no restarts (yet), fail when data transfer stalls


aarshkshah1992 (approved):
LGTM !

PR #437, opened 2020-10-14
Fix panil on transfer id not present

hannahhoward:
For legacy deals in transferring state, we did not record the restart ID for the transfer so we can't restart them. So we need to fail the deal.

Right now we are just trying to restart anyway which causes a nil panic -- eek!


arajasek (approved):
Makes total sense

PR #438, opened 2020-10-19
Preserve MaxPieceSize when calling SetAsk

ingar:
## Problem
If `SetAsk()` is called without a parameter to configure `MaxPieceSize`, it will revert to the default value.

## Solution
If a previously-stored ask is available, use its `MaxPieceSize` as a default, like is done for `MinPieceSize`


hannahhoward (approved):
LGTM

PR #440, opened 2020-11-03
Remove temp files

hannahhoward:
# Goals

Analysis of code indicates one possible way for stray temp files to be left hanging around is when commP's do not match. In this case, temporary piece files are generated but not recorded on the deal, and are therefore not cleaned up in the StorageDealFailing state.

# Implementation

Save the piecePath and filestorePath when commP does not match so that the files get deleted


ingar (approved):
Makes sense!

PR #449, opened 2020-11-17
Remove temp files

hannahhoward:
Do not generate temporary files for pieces in normal workflow


hannahhoward:
I believe the test failure on ClientDealAccepted is real, most likely due to some reduced resumability for the code in the HandoffDeal function. I suspect that the problem has to do deleting the blockstore twice. (the blockstore now needs to hang around through on deal handed off? Not sure)

dirkmc:
@hannahhoward I moved deletion of the deal store to `CleanupDeal()` (where the piece path and metadata path are cleaned up) and this fixed the test failures in `TestRestartClient/ClientEventDealAccepted`.
I believe it makes sense for deal store deletion to happen there because the client may want to retrieve the data before it's been sealed, right?

dirkmc (approved):
LGTM 👍

PR #450, opened 2020-11-18
Fail fast in tests that use a waitgroup Wait()

dirkmc:
We have several tests that use waitgroup.Wait(), which can lock up the test until it hits the generic test timeout.
This change makes waitgroup.Wait() respect the context.


hannahhoward (approved):
LGTM

PR #453, opened 2020-11-20
Add StorageDealAwaitingPreCommit state

dirkmc:
Markets side of https://github.com/filecoin-project/lotus/issues/4900


storagemarket/nodes.go
```diff
@@ -16,6 +16,11 @@ import (
 	"github.com/filecoin-project/go-fil-markets/shared"
 )
 
+// DealSectorCommittedCallback is a callback that runs when a sector is pre-committed
```
nonsense, line 19:
```suggestion
// DealSectorPreCommittedCallback is a callback that runs when a sector is pre-committed
```

nonsense:
I am still very new to this code, but overall looks good to me!

nonsense (approved):

hannahhoward (changes requested):
It is actually important to move the new deal state to the bottom of the list -- otherwise you need to write a migration for deal states. It's no worries -- I'll do it for you and merge, just a heads up.

nonsense:
> It is actually important to move the new deal state to the bottom of the list -- otherwise you need to write a migration for deal states.

Right, this was going to offset a lot of states, including `Active`, so we would have messed up the state of the client with respect to pretty much all deals. Good catch!

hannahhoward (approved):

PR #457, opened 2020-11-27
Cancel transfer should cause Storage deal to fail

dirkmc:
Depends on https://github.com/filecoin-project/go-data-transfer/pull/119 https://github.com/filecoin-project/go-data-transfer/pull/120
Addresses https://github.com/filecoin-project/lotus/issues/4997

This PR modifies the client and provider FSMs to transition to "deal failing" when the data transfer for the deal is cancelled.

If there is no network connectivity, the other party in the data transfer will not detect the deal cancellation (because it's not possible for the party that cancels the transfer to send a "cancel" message across the wire).

We should fix this in a follow-up PR.

Additionally we should ensure cancel transfer causes retrieval deals to fail in a follow-up PR. Issue: https://github.com/filecoin-project/lotus/issues/5034


hannahhoward (approved):
LGTM

PR #464, opened 2020-12-02
Normalize final states for cancelled retrieval deals

ingar:
## Problem
A retrieval deal can be cancelled by either the client or provider.  The deals, when cancelled, were either going to error states or cancelled states depending on if you were a client or provider.  In addition, this state transition was done without cleaning up the tracking, closing channels, etc.

## Solution
Make all deals that are "cancelled" by either the client or provider have a final state of `DealStatusCancelled`, with the reason for the cancel stored in the `msg` field.

- [x] Depends on https://github.com/filecoin-project/go-data-transfer/pull/123


retrievalmarket/dealstatus.go
```diff
@@ -45,10 +45,10 @@ const (
 	// continue processing the deal
 	DealStatusFundsNeeded
 
-	// DealStatusSendFunds indicats the client is now going to send funds because we reached the threshold of the last payment
+	// DealStatusSendFunds indicates the client is now going to send funds because we reached the threshold of the last payment
 	DealStatusSendFunds
 
-	// DealStatusSendFundsLastPayment indicats the client is now going to send final funds because
```
dirkmc, line 51: indicats! 🐱

retrievalmarket/impl/clientstates/client_fsm.go
```diff
@@ -270,18 +270,18 @@ var ClientEvents = fsm.Events{
...
 			}
 			return nil
 		},
 	),
 
-	// user manually cancells retrieval
+	// user manually cancels retrieval
 	fsm.Event(rm.ClientEventCancel).FromAny().To(rm.DealStatusCancelling).Action(func(deal *rm.ClientDealState) error {
-		deal.Message = "Retrieval Cancelled"
+		deal.Message = "Client cancelled retrieval"
```
dirkmc, line 284: I think that the provider can also cancel the deal - are we guaranteed to only get to this action if the client cancels?

dirkmc (approved):
One question but looks great 🎉

retrievalmarket/impl/clientstates/client_fsm.go
ingar: Yeah, when the provider cancels it comes from the data transfer event, translated into the `ClientEventProviderCancelled` event.

PR #480, opened 2021-02-01
retrieval: skip payment channel creation if total price is zero

dirkmc:


docs/retrievalclient.mmd
```diff
@@ -63,6 +63,7 @@ stateDiagram-v2
 	4 --> 8 : ClientEventPaymentChannelErrored
 	5 --> 8 : ClientEventPaymentChannelErrored
 	6 --> 8 : ClientEventPaymentChannelErrored
+	6 --> 13 : <invalid Value>
```
nonsense, line 66: Not sure how these `.mmd` files are generated, but this seems wrong, or?

dirkmc: Ah good catch, I forgot to add the stringified versoin

nonsense (approved):

PR #491, opened 2021-02-19
price specification by user should match what miner has in their ask

whyrusleeping:
This was a poor bit of UX that required users to multiply their storage price by the padded size of their data in order to propose the correct thing to the miner.

price should always be specified in the same terms. Miners asks say price in FIL/GB/Epoch, and what the user types should match this.

### DIFF
--- a/storagemarket/impl/client.go
+++ b/storagemarket/impl/client.go
@@ -338,6 +338,8 @@ func (c *Client) ProposeStorageDeal(ctx context.Context, params storagemarket.Pr
 		return nil, xerrors.Errorf("computing commP failed: %w", err)
 	}
 
+	pricePerEpoch := big.Mul(big.NewInt(int64(pieceSize.Padded())), params.Price)
+
 	if uint64(pieceSize.Padded()) > params.Info.SectorSize {
 		return nil, fmt.Errorf("cannot propose a deal whose piece size (%d) is greater than sector size (%d)", pieceSize.Padded(), params.Info.SectorSize)
 	}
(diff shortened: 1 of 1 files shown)


arajasek (approved):
SGWM

whyrusleeping:
What helptext? 
```
why@sirius ~/c/lotus (master) [1]> ./lotus client deal --help
NAME:
   lotus client deal - Initialize storage deal with a miner

USAGE:
   lotus client deal [command options] [dataCid miner price duration]

CATEGORY:
   STORAGE

OPTIONS:
   --manual-piece-cid value     manually specify piece commitment for data (dataCid must be to a car file)
   --manual-piece-size value    if manually specifying piece cid, used to specify size (dataCid must be to a car file) (default: 0)
   --from value                 specify address to fund the deal with
   --start-epoch value          specify the epoch that the deal should start at (default: -1)
   --fast-retrieval             indicates that data should be available for fast retrieval (default: true)
   --verified-deal              indicate that the deal counts towards verified client total (default: true if client is verified, false otherwise)
   --provider-collateral value  specify the requested provider collateral the miner should put up
   --help, -h                   show help (default: false)
```

arajasek:
The imaginary helptext that would explain what `price` means. Obviously.

dirkmc:
Reverted until we can figure out backwards compatibility:
https://github.com/filecoin-project/lotus/pull/5629#pullrequestreview-594085117

storagemarket/impl/client.go
```diff
@@ -338,6 +338,8 @@ func (c *Client) ProposeStorageDeal(ctx context.Context, params storagemarket.Pr
 		return nil, xerrors.Errorf("computing commP failed: %w", err)
 	}
 
+	pricePerEpoch := big.Mul(big.NewInt(int64(pieceSize.Padded())), params.Price)
```
magik6k, line 341: Pretty sure `params.Price` is per GB here

arajasek: based on how this gets into the actors proposal object, i don't think it is.

PR #493, opened 2021-02-19
Disallow concurrent retrieval deals for same peer/cid

dirkmc:
Addresses https://github.com/filecoin-project/lotus/issues/5443


nonsense (approved):
lgtm

retrievalmarket/impl/client.go
```diff
@@ -281,6 +296,31 @@ func (c *Client) Retrieve(ctx context.Context, payloadCID cid.Cid, params retrie
 	return dealID, nil
 }
 
+// Check if there's already an active retrieval deal with the same peer
+// for the same payload CID
+func (c *Client) checkForActiveDeal(payloadCID cid.Cid, pid peer.ID) error {
```
raulk, line 301: nit: It would be more idiomatic to return `(dealID, bool)` here, and create the error in the caller.

PR #495, opened 2021-02-23
Add funds to payment channel before allocating lane

dirkmc:
Before retrieving a file, the client must create a payment channel to pay for retrieval. Each deal has a separate lane in the payment channel that is used to manage funds for that particular deal.

The steps are:
1. Create payment channel with funds on chain
2. Wait for payment channel to be created
3. Create a payment channel lane for the deal
4. Transfer data

If a payment channel already exists in step 1, funds must be added to the existing channel. This PR fixes a bug where the retrieval market state machine would not wait for the add funds message to land on chain before continuing, which caused issues when creating vouchers against the payment channel.


retrievalmarket/impl/clientstates/client_fsm.go
```diff
@@ -105,8 +110,18 @@ var ClientEvents = fsm.Events{
...
 
+	// The payment channel add funds message has landed on chain
 	fsm.Event(rm.ClientEventPaymentChannelReady).
+		// If the payment channel between client and provider was being created
+		// for the first time, or if the payment channel had already been
+		// created for an earlier deal but the initial funding for this deal
+		// was being added, then we still need to allocate a payment channel
+		// lane
 		From(rm.DealStatusPaymentChannelCreating).To(rm.DealStatusPaymentChannelAllocatingLane).
+		From(rm.DealStatusPaymentChannelAddingInitialFunds).To(rm.DealStatusPaymentChannelAllocatingLane).
```
nonsense, line 121: I am a bit confused with naming here. Why `Initial` funds? If the payment channel existed and ran out of funds, wouldn't it be better to call it `AddingMoreFunds`?

nonsense: I guess it is `Initial`, because this is adding funds for a new deal, right?

nonsense (approved):

retrievalmarket/impl/clientstates/client_fsm.go
dirkmc: Right exactly, the funds are for the new deal 👍

PR #498, opened 2021-03-03
always try to return some message to the client

whyrusleeping:
This helps resolve the issues of getting tons of EOFs when trying to check on my deal


dirkmc (approved):
Thanks!

PR #507, opened 2021-03-22
close the reader after unsealing into blockstore

dirkmc:
There seems to be an issue where the error message when unsealing fails is getting swallowed: https://github.com/filecoin-project/lotus/issues/5829
To make sure any read errors are surfaced, this PR explicitly closes the reader after reading the the unsealed data into the blockstore and returns any error.


aarshkshah1992 (approved):
@dirkmc 

Closing the pipe will fail all further read and write operations. But, I'm not sure if this is the culprit of the problem we are seeing.

Anyways, this is better than we have and is the right thing to do.

PR #522, opened 2021-03-30
enable data rate monitoring for retrieval after payment channel create

dirkmc:
Depends on https://github.com/filecoin-project/go-data-transfer/pull/182

Currently the push or pull channel data rate monitoring starts automatically when the channel is opened.

However for retrieval we open a pull channel, then create a payment channel. It can take a long time to open a payment channel, which means that the data-rate will be zero during that time, and the pull channel monitor will try to restart the connection.

This PR enables data rate monitoring for retrieval only after the payment channel has been created.

TODO:
- [x] Enable data-rate monitoring after the payment channel has been created
- [ ] Add a mechanism to enable / disable auto-start of data rate monitoring separately for push vs pull channels
  The channel monitor configuration has a flag `EnableDataRateMonitoring` that indicates whether to start monitoring the data-rate as soon as the channel is opened. This flag applies to both push and pull channels.
  We want it to apply only for retrieval (pull channel).
- [ ] Add a mechanism to disable data-rate monitoring
  When the client runs out of funds, it tries to add more funds to the payment channel. In that case the client should stop monitoring the data-rate.
- [ ] Integration test that verifies the channel will not be restarted while waiting for the payment channel to be created
- [ ] Integration test that verifies the channel will not be restarted while waiting for the funds to be added to the payment channel

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/filecoin-project/go-address v0.0.3
 	github.com/filecoin-project/go-cbor-util v0.0.0-20191219014500-08c40a1e63a2
 	github.com/filecoin-project/go-commp-utils v0.0.0-20201119054358-b88f7a96a434
-	github.com/filecoin-project/go-data-transfer v1.4.0
+	github.com/filecoin-project/go-data-transfer v1.1.1-0.20210330094219-74b8bb238970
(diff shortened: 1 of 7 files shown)


dirkmc:
Superseded by https://github.com/filecoin-project/go-fil-markets/pull/540

PR #525, opened 2021-03-31
Flush out & fix retrieval bugs

aarshkshah1992:
Depends On https://github.com/filecoin-project/go-data-transfer/pull/186.

Closes https://github.com/filecoin-project/lotus/issues/5895.
Closes https://github.com/filecoin-project/go-fil-markets/issues/528.
Closes https://github.com/filecoin-project/go-fil-markets/issues/527.

### TODO
- [ ] Merge & Release https://github.com/filecoin-project/go-data-transfer/pull/186/commits
- [ ] Update deps

**Testing Proof**

```
go test -count=100 -run TestStorageRetrieval
PASS
ok  	github.com/filecoin-project/go-fil-markets/retrievalmarket	95.965s
```


dirkmc:
Looks great 👍 

Could we also add an integration test for this scenario where the unseal price is non-zero but the price per byte is zero 🙏

aarshkshah1992:
@dirkmc Can we merge this ?

The integration tests can then be written in the PR for https://github.com/filecoin-project/go-fil-markets/issues/528 as the tests will fail without that fix anyway.

retrievalmarket/impl/requestvalidation/requestvalidation.go
```diff
@@ -106,9 +106,10 @@ func (rv *ProviderRequestValidator) validatePull(receiver peer.ID, proposal *ret
 	}
 
 	pds := retrievalmarket.ProviderDealState{
-		DealProposal:   *proposal,
-		Receiver:       receiver,
-		LegacyProtocol: legacyProtocol,
+		DealProposal:    *proposal,
+		Receiver:        receiver,
+		LegacyProtocol:  legacyProtocol,
+		CurrentInterval: proposal.PaymentInterval,
```
aarshkshah1992, line 112: Not setting the `CurrentInterval ` here creates issues when we have a non zero price for Unsealing & bytes because we don't initialize this value correctly in the FSM when we jump directly to the `Give me money for the unsealing` state via the `RequestPayment` event instead of going via the `ProviderOpen` event .

retrievalmarket/impl/providerstates/provider_fsm.go
```diff
@@ -81,7 +81,11 @@ var ProviderEvents = fsm.Events{
 		From(rm.DealStatusFundsNeededUnseal).To(rm.DealStatusUnsealing).
 		Action(func(deal *rm.ProviderDealState, fundsReceived abi.TokenAmount) error {
 			deal.FundsReceived = big.Add(deal.FundsReceived, fundsReceived)
-			deal.CurrentInterval += deal.PaymentIntervalIncrease
+
+			// only update interval if the payment is for bytes and not for unsealing.
```
dirkmc, line 85: Nice catch 🎣

retrievalmarket/impl/requestvalidation/revalidator.go
```diff
@@ -178,12 +178,21 @@ func (pr *ProviderRevalidator) processPayment(dealID rm.ProviderDealIdentifier,
...
 	}
+
+	// This is hacky. But, if we don't do this, the data transfer module will resume data transfer
+	// even though we haven't finished reading the Unsealed data into the local Block-store since data transfer
+	// treats a "nil" error from here as a resumption signal.
+	if deal.Status == rm.DealStatusUnsealing || deal.Status == rm.DealStatusFundsNeededUnseal {
+		return nil, datatransfer.ErrNoOp
+	}
+
 	return nil, nil
```
dirkmc, line 196: IIUC in the worst case the provider may unpause too early, accidentally sending a bit of extra data to the client, but that's fine, this code _does_ guarantee that unpause only happens after unsealing has completed which is the important part 👍

dirkmc (approved):
Great work Aarsh! 🎉

dirkmc:
Note: I'm going to test this out on a live client / miner before merging

PR #531, opened 2021-04-02
add timeout for sending cancel message to peer when retrieval cancelled

dirkmc:


retrievalmarket/impl/integration_test.go
```diff
@@ -508,9 +508,7 @@ CurrentInterval: %d
 				t.FailNow()
 			case clientDealState = <-clientDealStateChan:
 			}
-			if testCase.failsUnseal {
-				assert.Equal(t, retrievalmarket.DealStatusErrored, clientDealState.Status)
-			} else if testCase.cancelled {
+			if testCase.failsUnseal || testCase.cancelled {
```
dirkmc, line 511: If there's an error unsealing on the provider, the best we can hope for on the client side is to receive a cancel message.
This test only passed before because the statemachine on the client was already shutdown when it received the Cancel event, so firing the event caused an error.
I fixed that bug so that the statemachine just ignores a Cancel event when the statemachine has already been shutdown

retrievalmarket/impl/provider_environments.go
```diff
@@ -21,6 +21,8 @@ import (
 	"github.com/filecoin-project/go-fil-markets/retrievalmarket/impl/dtutils"
 	"github.com/filecoin-project/go-fil-markets/retrievalmarket/impl/providerstates"
 	"github.com/filecoin-project/go-fil-markets/retrievalmarket/impl/requestvalidation"
+	"github.com/filecoin-project/go-fil-markets/shared"
+	mktshared "github.com/filecoin-project/go-fil-markets/shared"
```
nonsense, line 25: Nit, but we need only one of these.

PR #535, opened 2021-04-07
Allow anonymous submodule checkout

ribasushi:


dirkmc:
LGTM but could you add a couple of lines explaining what use case we need this for?

ribasushi:
Errr... if you do:
```
git clone https://github.com/filecoin-project/go-fil-markets
cd go-fil-markets
make build
```

it will fail unless you are a part of the `filecoin-project` org and have your ssh-agent at the ready to checkout the needed submodules.

dirkmc (approved):

PR #539, opened 2021-04-09
[WIP Draft] Retreival restart phase 1

aarshkshah1992:
Depends on https://github.com/filecoin-project/go-data-transfer/pull/193

PR #540, opened 2021-04-19
fix restarts during data transfer for a retrieval deal

### LINKED ISSUE #533: Integration test: retrieval deal restart during transfer
Create a test that
1. Starts a retrieval deal
2. Bounces the connection when the data transfer is ongoing
3. Verifies that all data is transferred successfully

dirkmc:
Fixes https://github.com/filecoin-project/go-fil-markets/issues/533
Supersedes https://github.com/filecoin-project/go-fil-markets/pull/539
Depends on https://github.com/filecoin-project/go-data-transfer/pull/197

TODO:
- [x] Clean up logging & commented out code
- [x] Handle case where transfer is restarted in `validatePull()`
- [x] Manual test


retrievalmarket/impl/clientstates/client_fsm.go
```diff
@@ -214,10 +221,12 @@ var ClientEvents = fsm.Events{
 			rm.DealStatusFundsNeededLastPayment,
 			rm.DealStatusCheckComplete,
 			rm.DealStatusClientWaitingForLastBlocks).ToNoChange().
+		FromMany(rm.DealStatusSendFunds, rm.DealStatusSendFundsLastPayment).To(rm.DealStatusOngoing).
 		FromMany(paymentChannelCreationStates...).ToJustRecord().
 		Action(recordReceived),
 
 	fsm.Event(rm.ClientEventSendFunds).
+		FromMany(rm.DealStatusSendFunds, rm.DealStatusSendFundsLastPayment).To(rm.DealStatusOngoing).
```
aarshkshah1992, line 229: Why this particular transition ? So, we've processed a payment request and decided we need to send funds.

Now, if we were already in a state of sending funds, we are moving out of it. Why are we moving out of it ? 

1. What will go wrong if we don't move out of it ?
2. What about the fact that we do need to send money for this  `ClientEventSendFunds ` event ? Are you betting that going to the `Ongoing` state will again trigger a request to send money as I see that we do check for remaining payments after moving to the `Ongoing` state.

retrievalmarket/impl/clientstates/client_fsm.go
```diff
@@ -252,35 +261,68 @@ var ClientEvents = fsm.Events{
...
-			deal.UnsealFundsPaid = big.Add(deal.UnsealFundsPaid, paymentForUnsealing)
-
-			// If the price per bytes is zero, we ONLY need to account for the Unsealing payments here.
-			if !deal.PricePerByte.IsZero() {
-				bytesPaidFor := big.Div(big.Sub(deal.PaymentRequested, paymentForUnsealing), deal.PricePerByte).Uint64()
-				if bytesPaidFor >= deal.CurrentInterval {
-					deal.CurrentInterval += deal.DealProposal.PaymentIntervalIncrease
-				}
-				deal.BytesPaidFor += bytesPaidFor
+		Action(func(deal *rm.ClientDealState, voucherAmt abi.TokenAmount) error {
```
aarshkshah1992, line 281: great refactor here !

retrievalmarket/impl/clientstates/client_fsm.go
```diff
@@ -297,7 +339,8 @@ var ClientEvents = fsm.Events{
 	// should wait for the last blocks to arrive (only needed when price
 	// per byte is zero)
 	fsm.Event(rm.ClientEventWaitForLastBlocks).
-		From(rm.DealStatusCheckComplete).To(rm.DealStatusClientWaitingForLastBlocks),
+		From(rm.DealStatusCheckComplete).To(rm.DealStatusClientWaitingForLastBlocks).
+		From(rm.DealStatusCompleted).ToJustRecord(),
```
aarshkshah1992, line 343: assuming this is one of "those".

retrievalmarket/impl/clientstates/client_fsm.go
```diff
@@ -179,7 +181,10 @@ var ClientEvents = fsm.Events{
 		FromMany(
 			rm.DealStatusOngoing,
 			rm.DealStatusBlocksComplete,
-			rm.DealStatusFundsNeeded).To(rm.DealStatusFundsNeeded).
+			rm.DealStatusFundsNeeded,
+			rm.DealStatusFundsNeededLastPayment).To(rm.DealStatusFundsNeeded).
+		From(rm.DealStatusSendFunds).To(rm.DealStatusOngoing).
```
aarshkshah1992, line 186: So we are moving to `Ongoing` here because we might be stuck in the `DealStatusSendFunds ` state -> letting the Ongoing state handler check if we have any dues left and if we do -> that handler will trigger a `ClientEventPaymentRequested ` again ? 

That's the idea behind adding this particular transition, right ?

retrievalmarket/impl/clientstates/client_fsm.go
```diff
@@ -252,35 +261,68 @@ var ClientEvents = fsm.Events{
...
-				if bytesPaidFor >= deal.CurrentInterval {
-					deal.CurrentInterval += deal.DealProposal.PaymentIntervalIncrease
-				}
-				deal.BytesPaidFor += bytesPaidFor
+		Action(func(deal *rm.ClientDealState, voucherAmt abi.TokenAmount) error {
+			// Reduce the payment requested by the amount of funds sent.
+			// Note that it may not be reduced to zero, if a new payment
+			// request came in while this one was being processed.
+			sentAmt := big.Sub(voucherAmt, deal.FundsSpent)
+			deal.PaymentRequested = big.Sub(deal.PaymentRequested, sentAmt)
```
aarshkshah1992, line 286: What purpose does the `deal.PaymentRequested` variable serve on the client now ?

retrievalmarket/impl/clientstates/client_states.go
```diff
@@ -89,42 +92,97 @@ func Ongoing(ctx fsm.Context, environment ClientDealEnvironment, deal rm.ClientD
...
+		log.Debugf("client: payment needed: unseal price %d > unseal paid %d",
+			deal.UnsealPrice, deal.UnsealFundsPaid)
+		return ctx.Trigger(rm.ClientEventSendFunds)
+	}
+
+	// If all bytes received have been paid for, we don't need to send funds
+	if deal.BytesPaidFor >= deal.TotalReceived {
+		log.Debugf("client: no payment needed: bytes paid for %d >= bytes received %d",
+			deal.BytesPaidFor, deal.TotalReceived)
+		return nil
```
aarshkshah1992, line 106: Why not trigger a `ClientEventPaymentNotSent ` here to move out to the `Ongoing ` state ?

retrievalmarket/impl/clientstates/client_states.go
```diff
@@ -89,42 +92,97 @@ func Ongoing(ctx fsm.Context, environment ClientDealEnvironment, deal rm.ClientD
...
+	totalBytesToPayFor := deal.TotalReceived
+
+	// If unsealing has been paid for, and not all blocks have been received
+	if deal.UnsealFundsPaid.GreaterThanEqual(deal.UnsealPrice) && !deal.AllBlocksReceived {
+		// If the number of bytes received is less than the number required
+		// for the current payment interval, no need to send a payment
+		if totalBytesToPayFor < deal.CurrentInterval {
+			log.Debugf("client: ignoring payment request for %d: total bytes to pay for %d < interval %d",
+				deal.PaymentRequested, totalBytesToPayFor, deal.CurrentInterval)
+			return nil
```
aarshkshah1992, line 143: ctx.Trigger(rm.ClientEventPaymentNotSent) here ?

aarshkshah1992 (commented):
@dirkmc 

First round of comments/questions.

retrievalmarket/impl/dtutils/dtutils.go
```diff
@@ -89,18 +89,18 @@ func clientEventForResponse(response *rm.DealResponse) (rm.ClientEvent, []interf
 		return rm.ClientEventLastPaymentRequested, []interface{}{response.PaymentOwed}
 	case rm.DealStatusCompleted:
 		return rm.ClientEventComplete, nil
-	case rm.DealStatusFundsNeeded:
+	case rm.DealStatusFundsNeeded, rm.DealStatusOngoing:
```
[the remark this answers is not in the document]
dirkmc: When the provider needs funds, it goes into the "funds needed" state and sends the client a message asking for funds.
When the provider receives a payment it goes back to "ongoing".
With restarts sometimes the provider will send us a message when it's in the ongoing state, so this just covers that case.

retrievalmarket/impl/clientstates/client_states.go
dirkmc: Fixed 👍

retrievalmarket/impl/requestvalidation/requestvalidation.go
```diff
@@ -105,6 +106,12 @@ func (rv *ProviderRequestValidator) validatePull(receiver peer.ID, proposal *ret
 		return nil, errors.New("incorrect selector for this proposal")
 	}
 
+	// If the validation is for a restart request, the state is already being
+	// tracked so no further action is required
+	if isRestart {
+		return nil, nil
```
aarshkshah1992, line 112: because we are not returning a `ErrPause` here, the GS transfer will resume. 

If we restarted before Unsealing, this could fail.
If we restarted after asking for payment but before getting it, we'll wrongly resume here. 
etc etc.

retrievalmarket/impl/requestvalidation/revalidator.go
```diff
@@ -220,12 +250,21 @@ func (pr *ProviderRevalidator) OnPullDataSent(chid datatransfer.ChannelID, addit
 		return true, nil, err
 	}
 
+	// Calculate how much data has been sent in total
 	channel.totalSent += additionalBytesSent
-	if channel.pricePerByte.IsZero() || channel.totalSent-channel.totalPaidFor < channel.interval {
+	if channel.pricePerByte.IsZero() || channel.totalSent < channel.interval {
```
aarshkshah1992, line 255: Can you explain the accounting change in the interval that led to this change of checking total sent against the interval rather than checking the total unpaid for bytes against the interval ?

aarshkshah1992 (commented):
@dirkmc Have finished the review. Some questions/concerns.

Astounding work !

retrievalmarket/types.go
```diff
@@ -244,6 +256,28 @@ func (p Params) SelectorSpecified() bool {
...
+	var target uint64
+	for target < currentInterval {
+		lowerBound = target
+		target += intervalSize
+		intervalSize += p.PaymentIntervalIncrease
+	}
+	return lowerBound
+}
+
+func (p Params) NextInterval(currentInterval uint64) uint64 {
```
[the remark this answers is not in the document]
dirkmc: It's easier to keep track of the accounting if we count totals from zero, instead of counting deltas from some interval point - when there's a restart, using deltas gets much more complicated.
So we're now treating the interval as an absolute value from zero, instead of as a delta. eg:
```
Delta intervals:    1000, 1500, 2000
Absolute intervals: 1000, 2500, 4500
```

retrievalmarket/impl/providerstates/provider_fsm.go
```diff
@@ -58,6 +58,7 @@ var ProviderEvents = fsm.Events{
 	// request payment
```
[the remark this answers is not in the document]
dirkmc: Agreed, good catch 👍 
I will add some tests.

retrievalmarket/impl/providerstates/provider_fsm.go
```diff
@@ -79,12 +80,13 @@ var ProviderEvents = fsm.Events{
 		From(rm.DealStatusFundsNeeded).To(rm.DealStatusOngoing).
 		From(rm.DealStatusFundsNeededLastPayment).To(rm.DealStatusFinalizing).
 		From(rm.DealStatusFundsNeededUnseal).To(rm.DealStatusUnsealing).
+		FromMany(rm.DealStatusBlocksComplete, rm.DealStatusOngoing, rm.DealStatusFinalizing).ToJustRecord().
```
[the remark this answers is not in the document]
dirkmc: It's possible for the client to send the same voucher twice. Note that the voucher always updates the value (it's not a delta). So for example if the client sends a voucher for 0.5 FIL, then 0.1 FIL then another for 0.1 FIL, the total value of funds received is 0.1 FIL.
If we assume that the system can handle the same voucher being sent twice it makes the accounting much simpler, and after a restart it's safe for the client to resend its last voucher to ensure that both sides are in sync.

retrievalmarket/impl/requestvalidation/requestvalidation.go
dirkmc: So I think that actually returning nil here will not explicitly resume the data transfer:
https://github.com/filecoin-project/go-data-transfer/blob/a175afc88c43158de8b26982e48d9f50b24474d9/impl/receiver.go#L30

retrievalmarket/impl/requestvalidation/revalidator.go
dirkmc: I added some detail in [this comment](https://github.com/filecoin-project/go-fil-markets/pull/540#discussion_r620131347)
Basically we're treating the interval as an absolute value instead of a delta now.

retrievalmarket/impl/requestvalidation/requestvalidation.go
aarshkshah1992: @dirkmc 

Yes, but the transfer is not explicitly paused here.

This hook is called when Graphsync receives a new Pull request. If we don't explicitly ask Graphsync to pause this new request, it will continue with the transfer. That is why `ValidatePull` currently returns a Pause.

IIUC, returning an `ErrResume` makes sense only when things have already been paused . There's no point in asking GS to resume a req that hasn't been paued.

retrievalmarket/impl/providerstates/provider_fsm.go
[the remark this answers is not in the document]
dirkmc: Tests added (see `TestBounceConnectionDealTransferUnsealing`)

retrievalmarket/impl/provider_environments.go
```diff
@@ -217,7 +217,8 @@ type providerStoreGetter struct {
 
 func (psg *providerStoreGetter) Get(otherPeer peer.ID, dealID retrievalmarket.DealID) (*multistore.Store, error) {
 	var deal retrievalmarket.ProviderDealState
-	err := psg.p.stateMachines.GetSync(context.TODO(), retrievalmarket.ProviderDealIdentifier{Receiver: otherPeer, DealID: dealID}, &deal)
+	provDealID := retrievalmarket.ProviderDealIdentifier{Receiver: otherPeer, DealID: dealID}
+	err := psg.p.stateMachines.Get(provDealID).Get(&deal)
```
aarshkshah1992, line 221: Why have we replaced `GetSync` with `Get` here ?

aarshkshah1992 (approved):
Just one comment. LGTM.

retrievalmarket/impl/provider_environments.go
dirkmc: Good question, it was a bug that took me a while to track down.
While the deal moves to the unsealing state, the state machine calls a method to unseal and blocks until it returns. So no other events can be processed until unsealing completes.
If there is a restart, the client creates a new graphsync request, and on the response side, the provider needs to attach a datastore to the graphsync response. It gets the datastore by calling `providerStoreGetter.Get` (the method above). However if we call stateMachines.GetSync it will be blocked until unsealing complete, and the graphsync response times out, so the deal fails. We only need to access the state machine to get the StoreID, which is set when the state machine is initiated, so there's no need to use stateMachines.GetSync, we can just call stateMachines.Get, which doesn't get blocked.

PR #541, opened 2021-04-20
Test Retrieval for offline deals

aarshkshah1992:
@dirkmc This looks good on a 100 runs.


dirkmc (commented):
Looks good, thanks 👍 

It looks like there's some code that is shared with TestStorageRetrieval that we can probably extract into a function.

PR #542, opened 2021-04-22
Dynamic Retrieval Pricing

aarshkshah1992:
Markets PR for the Dynamic Retrieval Pricing feature. Please see the corresponding [Lotus PR](https://github.com/filecoin-project/lotus/pull/6175#issue-629111010) for the motivation behind this work.

Here is an explanation of the changes made in the Markets module to support Dynamic Retrieval Pricing:
-  Instead of using the price configured in the `Ask Store` to blanket price all Retrieval deals, we now call the Pricing function configured by Lotus with the following parameters and use the output of the function to price a retrieval deal:

```
type PricingInput struct {
	// PayloadCID is the cid of the payload to retrieve.
	PayloadCID cid.Cid
	// PieceCID is the cid of the Piece from which the Payload will be retrieved.
	PieceCID cid.Cid
	// PieceSize is the size of the Piece from which the payload will be retrieved.
	PieceSize abi.UnpaddedPieceSize
	// Client is the peerID of the retrieval client.
	Client peer.ID
	// VerifiedDeal is true if there exists a verified storage deal for the PayloadCID.
	VerifiedDeal bool
	// Unsealed is true if there exists an unsealed sector from which we can retrieve the given payload.
	Unsealed bool
	// CurrentAsk is the current configured ask in the ask-store by using the lotus `set-ask` retrieval 
	// CLI command.
	CurrentAsk Ask
}
```

- Introduces/Exports a "default pricing function" that Lotus can configure Markets to use for retrieval pricing. This pricing function:
  *  Does NOT charge for Unsealing if we already have an Unsealed copy of a sector file containing the retrieval payload.
  *  Does NOT charge for data-transfer if there exists a Verified storage deal containing the retrieval payload AND the user hasn't turned off this particular functionality in the [Lotus config](https://github.com/filecoin-project/lotus/blob/1b5c1f0588399a8153c7d535bae795c2530116a0/node/config/def.go#L99).
  
- If the Miner has multiple Sector files containing the required retrieval payload AND the client hasn't mandated that we retrieve the payload from a _specific_ Piece, we will ALWAYS read the payload from an Unsealed Sector file and also price deals accordingly.

- If the client wants to retrieve the payload from a _specific_ Piece,  the above logic still applies but we will only consider Sector files that have the Piece mandated by the user.


retrievalmarket/impl/provider.go
```diff
@@ -51,6 +54,7 @@ type Provider struct {
 	dealDecider          DealDecider
 	askStore             retrievalmarket.AskStore
 	disableNewDeals      bool
+	dealPricingFunc      DealPricingFunc
```
raulk, line 57: "Deal" is too general a term. Let's remove the ambiguity here.

```suggestion
	retrievalPricingFunc      RetrievalPricingFunc
```

retrievalmarket/impl/provider.go
```diff
@@ -305,13 +313,48 @@ func (p *Provider) HandleQueryStream(stream rmnet.RetrievalQueryStream) {
...
+					storageDeals = append(storageDeals, d.DealID)
+				}
+			} else {
+				// If the user does NOT want to retrieve the payload from a specific piece, we'll have to inspect all deals
+				// made for that piece to quote a price for retrieval.
+				storageDeals, err = getAllDealsContainingPayload(p.pieceStore, query.PayloadCID)
+				if err != nil {
+					log.Errorf("Retrieval query: getAllDealsContainingPayload: %s", err)
+					return
+				}
```
dirkmc, line 336: It seems like in a few places in this function, if there is an error we log something but don't send QueryResponseError to the client. Can we restructure the code so that the client gets a response with an error message?
eg see the way the error response to GetMinerWorkerAddress() is handled above

retrievalmarket/impl/provider.go
```diff
@@ -327,6 +370,31 @@ func (p *Provider) HandleQueryStream(stream rmnet.RetrievalQueryStream) {
 	}
 }
 
+func (p *Provider) GetDynamicAsk(ctx context.Context, input retrievalmarket.PricingInput, storageDeals []abi.DealID) (retrievalmarket.Ask, error) {
```
dirkmc, line 373: Could we add a comment to GetDynamicAsk explaining the difference between this function and GetAsk

retrievalmarket/impl/provider.go
```diff
@@ -327,6 +370,31 @@ func (p *Provider) HandleQueryStream(stream rmnet.RetrievalQueryStream) {
...
+	dp, err := p.node.GetRetrievalPricingInput(ctx, input.PieceCID, storageDeals)
+	if err != nil {
+		return retrievalmarket.Ask{}, xerrors.Errorf("GetDealPricingParams: %s", err)
```
dirkmc, line 376:
```suggestion
		return retrievalmarket.Ask{}, xerrors.Errorf("GetRetrievalPricingInput: %s", err)
```

retrievalmarket/impl/provider_environments.go
```diff
@@ -30,17 +30,50 @@ type providerValidationEnvironment struct {
 	p *Provider
 }
 
-func (pve *providerValidationEnvironment) GetPiece(c cid.Cid, pieceCID *cid.Cid) (piecestore.PieceInfo, error) {
+func (pve *providerValidationEnvironment) GetAsk(ctx context.Context, payloadCid cid.Cid, pieceCid *cid.Cid,
```
dirkmc, line 33: It seems like all the code in this function is shared with some code in Provider.HandleQueryStream()
Can we refactor to keep the common code in one place?

retrievalmarket/impl/provider_environments.go
```diff
@@ -188,25 +221,87 @@ func (pde *providerDealEnvironment) CloseDataTransfer(ctx context.Context, chid
 func (pde *providerDealEnvironment) DeleteStore(storeID multistore.StoreID) error {
 	return pde.p.multiStore.Delete(storeID)
 }
-func getPieceInfoFromCid(pieceStore piecestore.PieceStore, payloadCID, pieceCID cid.Cid) (piecestore.PieceInfo, error) {
+
+func pieceInUnsealedSector(ctx context.Context, n retrievalmarket.RetrievalProviderNode, pieceInfo piecestore.PieceInfo) bool {
+	for _, di := range pieceInfo.Deals {
+		isUnsealed, err := n.IsUnsealed(ctx, di.SectorID, di.Offset.Unpadded(), di.Length.Unpadded())
+		if err != nil {
+			continue
```
dirkmc, line 229: Does it make sense to log the error here?

retrievalmarket/impl/provider_environments.go
```diff
@@ -188,25 +221,87 @@ func (pde *providerDealEnvironment) CloseDataTransfer(ctx context.Context, chid
...
+		}
+		for _, d := range pieceInfo.Deals {
+			dealsIds = append(dealsIds, d.DealID)
+		}
+	}
+
+	if len(dealsIds) == 0 {
+		return nil, xerrors.New("no deals found")
+	}
```
dirkmc, line 258: If there was an error calling GetPieceInfo, and no other call to GetPieceInfo is successful, let's return the last error

retrievalmarket/impl/provider_environments.go
```diff
@@ -188,25 +221,87 @@ func (pde *providerDealEnvironment) CloseDataTransfer(ctx context.Context, chid
...
-		if err == nil {
-			if pieceCID.Equals(cid.Undef) || pieceInfo.PieceCID.Equals(pieceCID) {
-				return pieceInfo, nil
+		if err != nil {
+			lastErr = err
+			continue
+		}
+
+		// if client wants to retrieve the payload from a specific piece, just return that piece.
+		if !pieceCID.Equals(cid.Undef) && pieceInfo.PieceCID.Equals(pieceCID) {
```
dirkmc, line 279:
```suggestion
		if pieceCID.Defined() && pieceInfo.PieceCID.Equals(pieceCID) {
```

retrievalmarket/impl/provider_test.go
```diff
@@ -37,32 +37,522 @@ import (
 	tut "github.com/filecoin-project/go-fil-markets/shared_testutil"
 )
 
+func TestDynamicPricing(t *testing.T) {
```
dirkmc, line 40: I love how thorough this test is 👍 👍

retrievalmarket/impl/provider_test.go
```diff
@@ -77,46 +567,88 @@ func TestHandleQueryStream(t *testing.T) {
...
 				PieceCIDFound: retrievalmarket.QueryItemAvailable,
 				Size:          expectedSize,
 			},
+			expectedPricePerByte:            expectedPricePerByte,
+			expectedPaymentInterval:         expectedPaymentInterval,
+			expectedPaymentIntervalIncrease: expectedPaymentIntervalIncrease,
+			expectedUnsealPrice:             expectedUnsealPrice,
+		},
+
+		{name: "When PieceCID is not provided, prefer a piece for which an unsealed sector already exists and price it accordingly",
```
dirkmc, line 630: 👍 👍

retrievalmarket/impl/providerstates/provider_states.go
```diff
@@ -30,7 +30,23 @@ type ProviderDealEnvironment interface {
 }
 
 func firstSuccessfulUnseal(ctx context.Context, node rm.RetrievalProviderNode, pieceInfo piecestore.PieceInfo) (io.ReadCloser, error) {
+
+	// prefer an unsealed sector containing the piece if one exists
+	for _, deal := range pieceInfo.Deals {
+		isUnsealed, err := node.IsUnsealed(ctx, deal.SectorID, deal.Offset.Unpadded(), deal.Length.Unpadded())
+		if err != nil {
+			continue
```
dirkmc, line 38: Let's return the last error if it's not possible to find any sectors with the piece

retrievalmarket/impl/providerstates/provider_states.go
```diff
@@ -30,7 +30,23 @@ type ProviderDealEnvironment interface {
...
+		}
+		if isUnsealed {
+			reader, err := node.UnsealSector(ctx, deal.SectorID, deal.Offset.Unpadded(), deal.Length.Unpadded())
```
dirkmc, line 41: I didn't get this the first time I looked at it - let's add a comment explaining that UnsealSector will get the data from the cache if it's already unsealed

retrievalmarket/impl/requestvalidation/requestvalidation.go
```diff
@@ -137,9 +139,22 @@ func (rv *ProviderRequestValidator) validatePull(receiver peer.ID, proposal *ret
...
 func (rv *ProviderRequestValidator) acceptDeal(deal *retrievalmarket.ProviderDealState) (retrievalmarket.DealStatus, error) {
+	pieceInfo, isUnsealed, err := rv.env.GetPiece(deal.PayloadCID, deal.PieceCID)
+	if err != nil {
+		if err == retrievalmarket.ErrNotFound {
+			return retrievalmarket.DealStatusDealNotFound, err
+		}
+		return retrievalmarket.DealStatusErrored, err
+	}
+
+	ask, err := rv.env.GetAsk(context.TODO(), deal.PayloadCID, deal.PieceCID, pieceInfo, isUnsealed, deal.Receiver)
```
dirkmc, line 150: Is it possible to use a context that is tied into the life cycle of the request? It seems like this call could hang forever.

dirkmc (commented):
The test coverage is really nice 👍

retrievalmarket/impl/provider.go
```diff
@@ -274,57 +288,120 @@ func (p *Provider) HandleQueryStream(stream rmnet.RetrievalQueryStream) {
...
+
+	dp.PayloadCID = input.PayloadCID
+	dp.PieceCID = input.PieceCID
+	dp.Unsealed = input.Unsealed
+	dp.Client = input.Client
+	dp.CurrentAsk = *currAsk
+
+	ask, err := p.retrievalPricingFunc(ctx, dp)
+	if err != nil {
+		return retrievalmarket.Ask{}, xerrors.Errorf("dealPricingFunc: %s", err)
```
dirkmc, line 402:
```suggestion
		return retrievalmarket.Ask{}, xerrors.Errorf("retrievalPricingFunc: %w", err)
```

retrievalmarket/impl/provider_environments.go
```diff
@@ -30,17 +30,38 @@ type providerValidationEnvironment struct {
...
+
+func (pve *providerValidationEnvironment) GetPiece(c cid.Cid, pieceCID *cid.Cid) (piecestore.PieceInfo, bool, error) {
 	inPieceCid := cid.Undef
 	if pieceCID != nil {
 		inPieceCid = *pieceCID
 	}
-	return getPieceInfoFromCid(pve.p.pieceStore, c, inPieceCid)
+
+	pi, isUnsealed, err := getPieceInfoFromCid(context.TODO(), pve.p.node, pve.p.pieceStore, c, inPieceCid)
+	return pi, isUnsealed, err
```
dirkmc, line 60:
```suggestion
	return getPieceInfoFromCid(context.TODO(), pve.p.node, pve.p.pieceStore, c, inPieceCid)
```

retrievalmarket/impl/provider_environments.go
```diff
@@ -188,25 +209,120 @@ func (pde *providerDealEnvironment) CloseDataTransfer(ctx context.Context, chid
 func (pde *providerDealEnvironment) DeleteStore(storeID multistore.StoreID) error {
 	return pde.p.multiStore.Delete(storeID)
 }
-func getPieceInfoFromCid(pieceStore piecestore.PieceStore, payloadCID, pieceCID cid.Cid) (piecestore.PieceInfo, error) {
+
+func pieceInUnsealedSector(ctx context.Context, n retrievalmarket.RetrievalProviderNode, pieceInfo piecestore.PieceInfo) bool {
+	for _, di := range pieceInfo.Deals {
+		isUnsealed, err := n.IsUnsealed(ctx, di.SectorID, di.Offset.Unpadded(), di.Length.Unpadded())
+		if err != nil {
+			log.Errorf("failed to find out is sector %s is unsealed, err=%s", di.SectorID, err)
```
dirkmc, line 217:
```suggestion
			log.Errorf("failed to find out if sector %d is unsealed, err=%s", di.SectorID, err)
```

retrievalmarket/impl/provider_environments.go
```diff
@@ -188,25 +209,120 @@ func (pde *providerDealEnvironment) CloseDataTransfer(ctx context.Context, chid
...
+		//  we only need to inspect storage deals made for that piece to quote a price.
+		for _, d := range pieceInfo.Deals {
+			storageDeals = append(storageDeals, d.DealID)
+		}
+	} else {
+		// If the user does NOT want to retrieve from a specific piece, we'll have to inspect all storage deals
+		// made for that piece to quote a price.
+		storageDeals, err = getAllDealsContainingPayload(pieceStore, payloadCID)
+		if err != nil {
+			return nil, xerrors.Errorf("failed to fetch deals for payload, err=%s", err)
```
dirkmc, line 242:
```suggestion
			return nil, xerrors.Errorf("failed to fetch deals for payload: %w", err)
```

retrievalmarket/impl/provider_environments.go
```diff
@@ -188,25 +209,120 @@ func (pde *providerDealEnvironment) CloseDataTransfer(ctx context.Context, chid
...
+			dealsIds = append(dealsIds, d.DealID)
+		}
+	}
+
+	if lastErr == nil && len(dealsIds) == 0 {
+		return nil, xerrors.New("no deals found")
+	}
+
+	if lastErr != nil && len(dealsIds) == 0 {
+		return nil, xerrors.Errorf("failed to fetch deals containing payload, err=%s", lastErr)
```
dirkmc, line 277:
```suggestion
		return nil, xerrors.Errorf("failed to fetch deals containing payload %s: %w", payloadCID, lastErr)
```

dirkmc (approved):

aarshkshah1992:
@dirkmc I've fixed a bug in commit `6c11597` wherein we were quoting the retrieval data transfer price based on the padded piece size instead of the unpadded piece size even though don't actually (obviously) transfer the padding across the wire.

raulk:
@aarshkshah1992 Mind updating this PR's description with an explanation of what has been done concretely in this module (as part of the linked Lotus side PR) + what's left, so that the communityi can follow?

Do we need to add more test coverage here?

aarshkshah1992:
@raulk @dirkmc 

This PR is in pretty good shape and has solid tests. I just need @dirkmc to look at one last small commit that I recently introduced:
https://github.com/filecoin-project/go-fil-markets/pull/542/commits/6c1159720a9b7bcedb0ee07b437adde022c607d6

If that looks good, we are good to go.

dirkmc:
Looks good, thanks Aarsh 👍 

Before we merge I'd like to do some real world testing, for correctness and also to make sure that it's performant.

retrievalmarket/types.go
```diff
@@ -416,3 +416,21 @@ type ChannelAvailableFunds struct {
...
+	PieceSize abi.UnpaddedPieceSize
+	// Client is the peerID of the retrieval client.
+	Client peer.ID
+	// VerifiedDeal is true if there exists a verified storage deal for the PayloadCID.
+	VerifiedDeal bool
+	// Unsealed is true if there exists an unsealed sector from which we can retrieve the given payload.
+	Unsealed bool
+	// CurrentAsk is the current configured ask in the ask-store.
+	CurrentAsk Ask
+}
```
ribasushi, line 436: ( I am well aware I am commenting on an old, merged PR )

@aarshkshah1992 @hannahhoward @willscott was it a deliberate decision to exclude the client-provided selector in this structure, or just an omission? Even if we do not have the tools for miners to interpret/decide on a selector-driven partial-retrieval, the structure is [being exposed to 3rd party tools](https://github.com/filecoin-project/lotus/discussions/6780#discussion-3466055). It is probably prudent to provide all the information available from the get-go, instead of adding it later on.

ribasushi: An extra nit regarding the title-comment:

> If the Miner has multiple Sector files containing the required retrieval payload AND the client hasn't mandated that we retrieve the payload from a specific Piece,

A specific deal `PayloadCID` (root-dag) + `PieceSize` will always invariably correspond to an identical `PieceCID`. That is - if a deal is stored in multiple sectors, some sealed and some unsealed, a `PieceCID` can not differentiate among them: it will be identical among them all.

raulk: Never too late to provide useful feedback! ;-)

Re: including the selector in the pricing function input: how useful would this be? The function can't _evaluate_ the selector against the data, and it can't really derive anything useful from it intrinsically AFAIK.

If it's to symbolise that a partial retrieval is being performed, for pricing purposes, the price-per-byte formulation of the quote already decouples the final price from the actual data being selected. We _could_ pass a "partial retrieval = true/false" flag, but it's not possible to calculate it correctly because the selector could be selecting the entire deal.

Re: extensibility/completeness, it's uninvasive to add a JSON field in the future if we do decide we want this for some reason. But until there's a standardised, canonical text representation for selectors, I don't think it's practical to even consider.

raulk: > A specific deal PayloadCID (root-dag) + PieceSize will always invariably correspond to an identical PieceCID. That is - if a deal is stored in multiple sectors, some sealed and some unsealed, a PieceCID can not differentiate among them: it will be identical among them all.

If there's an unsealed copy, we always select the unsealed copy as input to the pricing function, and as the source of the retrieval.

ribasushi: @raulk right, I was specifically questioning this part of the sentence:

> the client hasn't mandated that we retrieve the payload from a specific Piece

There is no way for a client to signal "I want to retrieve from this physical piece" via the struct in question: they all have the same `PieceCid`

raulk: Oh, what @aarshkshah1992 was trying to get to is that if "universal retrieval" is enabled (such that clients don't need to specify the PieceCID they want to retrieval a given PayloadCID from), and there are multiple PieceCIDs that contain that PayloadCID, we prefer the deal that's unsealed, should there be one. I don't think any of that implies that the client gets to choose the sector.

ribasushi: > But until there's a standardized, canonical text representation for selectors, I don't think it's practical to even consider.

The reason I brought this up is that the selector is already exposed in a [standard-ish encoded-cbor representation](https://github.com/filecoin-project/go-fil-markets/blob/v1.6.0-rc1/retrievalmarket/types.go#L245-L253) via the `RetrievalFilter` codepath: https://github.com/filecoin-project/lotus/blob/v1.11.0-rc2/markets/dealfilter/cli.go#L28-L62

I do not have a preference either way. Rather am raising the inconsistency of having two sets of external tools, one being fed the structure in this pr, the other being fed this: https://github.com/Murmuration-Labs/bitscreen/blob/master/testdata/retrieval_proposal.json

raulk: Thanks for the pointer. Note that this isn't a filtering function, but rather a pricing function. It runs when the retrieval provider is _asked_ for a quote. At that time, the selector is not available. So in terms of the data interface, the deal filter and the pricing function can't be symmetrical.

ribasushi: > It runs when the retrieval provider is asked for a quote. At that time, the selector is not available.

@raulk indeed you are correct: it is all commented out :cryingbear:
https://github.com/filecoin-project/go-fil-markets/blob/v1.6.0-rc1/retrievalmarket/types.go#L158-L167

PR #548, opened 2021-05-20
fix first make error

yaohcn:
Clone and first make all report error:
```sh
➜  go-fil-markets git:(fix-makefile) make all
/Library/Developer/CommandLineTools/usr/bin/make -C ./extern/filecoin-ffi/ .install-filcrypto
make[1]: *** No rule to make target `.install-filcrypto'.  Stop.
make: *** [.filecoin-build] Error 2
```


dirkmc (approved):
Thanks 👍

PR #549, opened 2021-05-31
For retrievals don't round down payments to interval lower bound

dirkmc:
Will fix https://github.com/filecoin-project/lotus/issues/6297 (when bubbled up to lotus)

When sending a payment voucher, we were rounding down the amount to the lower bound of the payment interval, so that payment would always proceed in exact intervals.
This PR changes the behaviour not to round down: the client will send a voucher that pays for all data that has been received (which may be more than the lower bound of the payment interval)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,37 @@
 # go-fil-markets changelog
 
+# go-fil-markets v1.5.0
+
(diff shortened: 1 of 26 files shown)


dirkmc:
Merged as part of https://github.com/filecoin-project/go-fil-markets/pull/550/files#

PR #556, opened 2021-06-15
[WIP for demo ONLY] Retrieval CAR Store usage demo

aarshkshah1992:


retrievalmarket/impl/providerstates/provider_states.go
```diff
@@ -58,18 +59,68 @@ func firstSuccessfulUnseal(ctx context.Context, node rm.RetrievalProviderNode, p
...
+			environment.Node(),
+			*deal.PieceInfo,
+		}); err != nil {
+			return err
+		}
+		// TODO We should remove the shard when the storage deal expires but that code probably belongs to the storage markets.
+	}
+
+	// get a read only blockstore that can be used to serve random access for the piece data here.
+	rb, err = ds.GetShardReadOnlyBlockstore(pieceCID)
```
dirkmc, line 97: It would be great if we could somehow call IsShardActive and ActivateShard from within GetShardReadOnlyBlockstore.
Otherwise it may be tricky to synchronize between different threads - eg two threads could end up unsealing the same data in parallel.

aarshkshah1992: @dirkmc I agree. The current design assumes there's only one Markets process.

PR #563, opened 2021-07-05
fix unpadded deal size

dirkmc:


aarshkshah1992:
The CAR team has agreed to preserve backward compat by having the Index/Blockstore generation code ignore the padding.

PR #564, opened 2021-07-06
Dagstore lotus mount Implementation with tests

aarshkshah1992:
### TODO

- [x] Unit Tests for the Loutus Mount API impl.
- [x] Replace mocked out DAG store with actual DAG store https://github.com/filecoin-project/go-fil-markets/pull/565
- [x] Register shard after handoff deal completes (when deal has been successfully sent for sealing)
- [ ] Destroy shard when storage deal expires / is slashed
  - [ ] `LotusMount.Stat.Exists` should be false when storage deal expires.
- [ ] Shard failure and recovery


dagstore/mount.go
```diff
@@ -2,81 +2,86 @@ package dagstore
...
 }
 
-func (m *mount) FetchUnsealedPiece(ctx context.Context, pieceCid cid.Cid) (io.ReadCloser, error) {
-	pieceInfo, err := m.pieceStore.GetPieceInfo(pieceCid)
+func NewLotusMount(pieceCid cid.Cid, api LotusMountAPI) (*LotusMount, error) {
+	u := fmt.Sprintf(lotusMountURL, lotusScheme, pieceCid.String())
+	url, err := url.Parse(u)
 	if err != nil {
-		return nil, xerrors.Errorf("failed to fetch pieceInfo: %w", err)
+		return nil, xerrors.Errorf("failed to parse URL, err=%s", err)
```
dirkmc, line 32:
```suggestion
		return nil, xerrors.Errorf("failed to parse URL: %w", err)
```

dagstore/mount.go
```diff
@@ -2,81 +2,86 @@ package dagstore
...
-	pieceInfo, err := m.pieceStore.GetPieceInfo(pieceCid)
+func (l *LotusMount) Close() error {
+	return nil
+}
+
+func (l *LotusMount) Stat(_ context.Context) (mount.Stat, error) {
+	size, err := l.Api.GetUnpaddedCARSize(l.PieceCid)
 	if err != nil {
-		return 0, xerrors.Errorf("failed to fetch pieceInfo, err=%w", err)
+		return mount.Stat{}, xerrors.Errorf("failed to fetch piece size, err=%s", err)
```
dirkmc, line 67:
```suggestion
		return mount.Stat{}, xerrors.Errorf("failed to fetch piece size: %w", err)
```

dirkmc (approved):
Couple of nits but LGTM 👍 

The mount implementation isn't really specific to lotus, should we call it something other than LotusMount?

aarshkshah1992:
@dirkmc Notes on integrating the Sharded DAG Store into Lotus and Markets:

1. Instantiate the Sharded DAG Store with the correct Config in Lotus. It's likely some of that Config will flow from the Lotus Config. The instantiation will take care of "rehydrating" the DAG Store so we don't need to worry about that. We'll instantiate and inject the Lotus Mount/Mount factory/Type here so they make their way into the Mount Registry of the DAG Store.

2. Register Shards in the storage market after deals have been successfully sent for sealing (i.e. storage deal enters the StorageDealAwaitingPrecommit state on the Miner). If this fails, keep retrying till this succeeds with the "Sharded DAG Store failure management and Recovery" mechanism. Note that we can pass in the CARv2 file that Graphsync wrote out for us as part of the "RegisterShard" call here in the RegisterOpts  option there to ask the Sharded DAG Store to simply scrap off the Index from the CARv2 file we already have rather than fetching the CARv1 payload from the Unsealer and indexing it all over again.
 
 3. In the retrieval Market, we'd call "AcquireShard" to get a ShardAccessor /ReadOnlyBlockstore that we can serve the retrieval from. It's important that we call Close  on this accessor when we no longer need it (successful or failed retrieval) so the DAG Store can clean up transient artefacts it creates for this retrieval. If the call "AcquireShard" fails, we'd have to use the same failure management/recovery mechanism mentioned above.

4. Also note that we might eventually decide that we need lazy registration of Shards (after having this discussion with @raulk in today's huddle) for all pre-existing deals to avoid the tricky migration effort there and in that case, the lazy shard registration i.e. calling ""RegisterShard" would also happen in the retrieval market.

aarshkshah1992:
@dirkmc Shifting review and TODOs to the `feat/dagstore-retrieval` here so I can review all the changes together.

PR #568, opened 2021-07-13
Sharded DAG Store failure Recovery and periodic GC

aarshkshah1992:
Uses the DAG Store Shard recovery functionality introduced in https://github.com/filecoin-project/dagstore/pull/60.


dagstore/dagstorewrapper.go
```diff
@@ -15,35 +19,91 @@ import (
...
+	}
+	failureCh := make(chan dagstore.ShardResult, 1)
+	dcfg := dagstore.Config{
+		TransientsDir: cfg.TransientsDir,
+		IndexDir:      cfg.IndexDir,
+		Datastore:     cfg.Datastore,
+		MountRegistry: registry,
+		FailureCh:     failureCh,
+	}
+	dagStore, err := dagstore.NewDAGStore(dcfg)
```
dirkmc, line 71: I think it's better to initialize the registry inside the wrapper like this 👍

dirkmc (commented):
LGTM 👍

PR #570, opened 2021-07-13
add migration to register deals as shards

dirkmc:


storagemarket/impl/shardregistration.go
```diff
@@ -0,0 +1,80 @@
...
+		var sealing bool
+		for _, state := range providerstates.ProviderSealingStates {
+			if deal.State == state {
+				sealing = true
+				break
+			}
+		}
+		if !sealing {
+			continue
+		}
```
raulk, line 58: What's the rationale for this? Will this support lazy initialization for sealed deals going forward?

storagemarket/impl/provider.go
```diff
@@ -581,25 +584,28 @@ func (p *Provider) start(ctx context.Context) error {
...
 	var deals []storagemarket.MinerDeal
-	err := p.deals.List(&deals)
+	err = p.deals.List(&deals)
 	if err != nil {
-		return err
+		return xerrors.Errorf("failed to fetch deals during startup: %w", err)
 	}
+	if err := p.shardReg.registerShards(ctx, deals); err != nil {
+		return fmt.Errorf("Failed to restart deals: %w", err)
+	}
```
raulk, line 595: We currently form a single batch. Is the intent to form multiple batches? Why do we need this batching? The DAG store will return from `RegisterShard` quickly.

storagemarket/impl/shardregistration.go
dirkmc: We only need to call RegisterShard for deals that have already passed the point at which the deal is handed off to the sealing subsystem.
If the deal has not yet reached that point it will be called from within the state machine once the deal reaches that point.

storagemarket/impl/provider.go
dirkmc: Makes sense, my assumption was that we needed to wait until there is a signal on the channel returned from RegisterShard before we can call the migration complete.
As discussed sync I will remove the batching as we only need to wait until RegisterShard returns (not until there is a signal on the channel)

storagemarket/impl/shardregistration.go
dirkmc: Added a comment explaining this in the code

storagemarket/impl/shardregistration.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+type ShardRegistration struct {
+	maddr       address.Address
+	ds          datastore.Datastore
+	dagStore    mktdagstore.DagStoreWrapper
+	sectorState SectorState
+}
+
+func NewShardRegistration(
```
raulk, line 36: Is there any special reasons the `ShardMigrator` should be here? Does it belong better in the Lotus side of things?

storagemarket/impl/shardregistration.go
```diff
@@ -0,0 +1,146 @@
...
+type SectorState interface {
+	StateSectorGetInfo(context.Context, address.Address, abi.SectorNumber, types.TipSetKey) (*miner.SectorOnChainInfo, error)
+	IsUnsealed(ctx context.Context, sector storage.SectorRef, offset storiface.UnpaddedByteIndex, size abi.UnpaddedPieceSize) (bool, error)
+}
+
+type ShardRegistration struct {
+	maddr       address.Address
+	ds          datastore.Datastore
+	dagStore    mktdagstore.DagStoreWrapper
+	sectorState SectorState
```
raulk, line 33: `SectorState` as a type name makes it sound like this is a sector's state. However, it's really an interface to _query sector states_. Idiomatic Go would recommend to name this something-er. `SectorStateQuerier`?

raulk (commented):
Partial review.

storagemarket/impl/shardregistration.go
```diff
@@ -0,0 +1,146 @@
...
+	// handed off to the sealing subsystem.
+	resch := make(chan dagstore.ShardResult, len(deals))
+	for _, deal := range deals {
+		if deal.Ref.PieceCid == nil {
+			continue
+		}
+
+		// Check if the deal has been handed off to the sealing subsystem
+		var sealing bool
+		for _, state := range providerstates.ProviderSealingStates {
```
raulk, line 74: I think the `sealing` term is a bit of a misnomer here, because this also includes deals that have finished sealing.

raulk: A more efficient way of doing this would be to form a `map[fsm.StateKey]struct{}` before the loop, and use an "exists" state to evaluate if the deal is in one of the required states.

storagemarket/impl/providerstates/provider_fsm.go
```diff
@@ -262,3 +262,13 @@ var ProviderFinalityStates = []fsm.StateKey{
...
+
+// ProviderSealingStates are the states on the happy path after hand-off to
+// the sealing subsystem
+var ProviderSealingStates = []fsm.StateKey{
+	storagemarket.StorageDealStaged,
+	storagemarket.StorageDealAwaitingPreCommit,
+	storagemarket.StorageDealSealing,
+	storagemarket.StorageDealFinalizing,
+	storagemarket.StorageDealActive,
+}
```
raulk, line 274:
```suggestion
var StatesKnownBySealingSubsystem = []fsm.StateKey{
	storagemarket.StorageDealStaged,
	storagemarket.StorageDealAwaitingPreCommit,
	storagemarket.StorageDealSealing,
	storagemarket.StorageDealFinalizing,
	storagemarket.StorageDealActive,
}
```

storagemarket/impl/providerstates/provider_fsm.go
```diff
@@ -262,3 +262,13 @@ var ProviderFinalityStates = []fsm.StateKey{
 	storagemarket.StorageDealSlashed,
 	storagemarket.StorageDealExpired,
 }
+
+// ProviderSealingStates are the states on the happy path after hand-off to
+// the sealing subsystem
+var ProviderSealingStates = []fsm.StateKey{
```
raulk, line 268: Are we 100% sure that no other deals need to be registered in the DAG store?

storagemarket/impl/shardregistration.go
```diff
@@ -0,0 +1,146 @@
...
+		r.dagStore.RegisterShardAsync(ctx, *deal.Ref.PieceCid, deal.CARv2FilePath, isUnsealed, resch)
+	}
+
+	// If there are any problems registering shards, just log an error
+	go func() {
+		for res := range resch {
+			if res.Error != nil {
+				log.Errorf("failed to register shard: %s", res.Error)
+			}
+		}
```
raulk, line 103: Better to start this goroutine should be started _before_ we perform the loop. Then we can get rid of the buffered channel and use an unbuffered one. (A buffered channel when there are, say 1MM deals, can get costly).

storagemarket/impl/shardregistration.go
```diff
@@ -0,0 +1,146 @@
...
+		// unsealed it will be initialized "lazily" once it's unsealed during
+		// retrieval)
+		r.dagStore.RegisterShardAsync(ctx, *deal.Ref.PieceCid, deal.CARv2FilePath, isUnsealed, resch)
+	}
+
+	// If there are any problems registering shards, just log an error
+	go func() {
+		for res := range resch {
+			if res.Error != nil {
+				log.Errorf("failed to register shard: %s", res.Error)
```
raulk, line 101:
```suggestion
				log.Warnf("dagstore migration: failed to register shard: %s", res.Error)
```

storagemarket/impl/shardregistration.go
```diff
@@ -0,0 +1,146 @@
...
+		// Register the deal as a shard with the DAG store, initializing the
+		// index immediately if the deal is unsealed (if the deal is not
+		// unsealed it will be initialized "lazily" once it's unsealed during
+		// retrieval)
+		r.dagStore.RegisterShardAsync(ctx, *deal.Ref.PieceCid, deal.CARv2FilePath, isUnsealed, resch)
+	}
+
+	// If there are any problems registering shards, just log an error
+	go func() {
+		for res := range resch {
```
raulk, line 99: This loop will never yield because the channel is never closed! We probably want to track the number of shards we registered, and consume that many times (with a for i := 0...).

storagemarket/impl/shardregistration.go
```diff
@@ -0,0 +1,146 @@
...
+			}
+		}
+	}()
+
+	// Completed registering all shards, so mark the migration as complete
+	err = r.ds.Put(shardRegKey, []byte{1})
+	if err != nil {
+		log.Errorf("failed to mark shards as registered: %s", err)
+	}
+
```
raulk, line 111: Should add a `ds.Sync()` here, to fsync the write.

storagemarket/impl/shardregistration.go
raulk: (So that the user knows this is coming from the migration goroutine).

storagemarket/impl/shardregistration.go
```diff
@@ -0,0 +1,146 @@
...
+			}
+		}
+		if !sealing {
+			continue
+		}
+
+		// Check if the deal is in an unsealed state
+		isUnsealed, err := r.isUnsealed(ctx, deal.SectorNumber)
+		if err != nil {
+			log.Errorf("failed to get unsealed state of deal with piece CID %s: %s", deal.Ref.PieceCid, err)
```
raulk, line 87: Since we're not in control of the value of `isUnsealed` on error, let's be defensive here and set this to `false`, because we want shards for which we couldn't determine their unsealed status to be registered as lazy.

storagemarket/impl/shardregistration.go
dirkmc: It needs to have knowledge of the deals state, and it start when markets starts, so I think the most natural fit is inside of markets.

storagemarket/impl/shardregistration.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func NewShardRegistration(
+	maddr address.Address,
+	ds datastore.Datastore,
+	dagStore mktdagstore.DagStoreWrapper,
+	sectorState SectorState,
+) *ShardRegistration {
+	return &ShardRegistration{
+		maddr:       maddr,
```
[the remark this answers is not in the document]
dirkmc: Changed to `providerAddr`

storagemarket/impl/shardregistration.go
dirkmc: Changed to `SectorStateAccessor`

storagemarket/impl/shardregistration.go
dirkmc: Changed `sealing` to `inSealingSubsystem`

It may be more efficient to use a map, however there are only 5 `ProviderSealingStates` and this is a one-off migration, so I think it's probably better to keep it simple at the cost of a few more milliseconds.

storagemarket/impl/providerstates/provider_fsm.go
dirkmc: I'm not sure I understand the question - are you asking about which are the states that deals requiring registration are in?

Basically it's any state after and including `StorageDealStaged` - This is the state in the new code in which `RegisterShard` is called.

The termination states are
- `storagemarket.StorageDealError`
- `storagemarket.StorageDealSlashed`
- `storagemarket.StorageDealExpired`
We don't want to register deals in any of these states.

So the remaining possibilities are:
- `storagemarket.StorageDealStaged`
- `storagemarket.StorageDealAwaitingPreCommit`
- `storagemarket.StorageDealSealing`
- `storagemarket.StorageDealFinalizing`
- `storagemarket.StorageDealActive`

You can examine [the state diagram](https://github.com/filecoin-project/go-fil-markets/blob/master/docs/storageprovider.mmd.png) to make sure.

storagemarket/impl/shardregistration.go
dirkmc: Good catch 👍

storagemarket/impl/shardregistration.go
dirkmc: I think somewhere we're going to have to keep a buffer:
- we could create a copy of the deals array and then filter in place, but that requires as much space as the original array
- we could create a new slice and append to it each time we register a deal but that is going to be much slower than reserving the buffer in advance
- we could run over the loop twice, which would save memory but at the cost of CPU

I don't think anyone has made a million deals, but even if they had the buffered channel would only require 1m pointers worth of space - so I think the buffered channel may be the least-bad solution here.

dagstore/dagstorewrapper.go
```diff
@@ -33,13 +33,26 @@ type MarketDAGStoreConfig struct {
 // the other parts of go-fil-markets
 type DagStoreWrapper interface {
 	// RegisterShard loads a CAR file into the DAG store and builds an index for it
-	RegisterShard(ctx context.Context, pieceCid cid.Cid, carPath string) error
+	RegisterShard(ctx context.Context, pieceCid cid.Cid, carPath string, eagerInit bool) error
+	// RegisterShardAsync loads a CAR file into the DAG store and builds an
+	// index for it, sending the result on the supplied channel on completion
+	RegisterShardAsync(ctx context.Context, pieceCid cid.Cid, carPath string, eagerInit bool, resch chan dagstore.ShardResult)
```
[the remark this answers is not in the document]
raulk: Yep, what I mean is that `RegisterShard` is really a _utility_ on top of `RegisterShardAsync` that adapts the async behaviour to sync behaviour. I don't think it deserves to be an interface member, as it's going to look the same across all implementations.

I'm proposing a simplification where you just have `RegisterShardAsync` in the interface (but renamed to `RegisterSync`), and a separate utility function (outside the interface).

```
func RegisterShardSync(ctx context.Context, dagStore DasStoreWrapper, pieceCid cid.Cid, carPath string, eagerInit bool)
```

that does the bridging?

dagstore/dagstorewrapper.go
```diff
@@ -174,9 +185,37 @@ func (ds *dagStoreWrapper) RegisterShard(ctx context.Context, pieceCid cid.Cid,
...
+}
+
+func (ds *dagStoreWrapper) RegisterShardAsync(ctx context.Context, pieceCid cid.Cid, carPath string, eagerInit bool, resch chan dagstore.ShardResult) {
+	err := ds.registerShardAsync(ctx, pieceCid, carPath, eagerInit, resch)
+	if err != nil {
+		select {
+		case <-ctx.Done():
+		case resch <- dagstore.ShardResult{Error: err}:
+		}
+	}
```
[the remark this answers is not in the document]
raulk: @dirkmc missed comment ^^

storagemarket/impl/shardregistration.go
raulk: If you start the goroutine before you start registering deals, you can make progress on the results consumption in parallel, and you can remove the channel buffer entirely, or keep it small. Right now you're using the channel buffer as a slice, and that's mildly inefficient.

storagemarket/impl/shardregistration.go
raulk: The double iteration is ugly :-( set membership check is in Go is very idiomatic as a map. If you don't want to form a map here, then just have the original set be a map?

storagemarket/impl/providerstates/provider_fsm.go
raulk: Thanks, that makes sense!

storagemarket/impl/shardregistration.go
dirkmc: The problem is that in order to know when to stop reading from the channel, you need to know the total number of deals. And you don't know that until the end of the for loop. So if you try to make progress on results consumption in parallel you're in a race between the go routine and the for loop.

dagstore/dagstorewrapper.go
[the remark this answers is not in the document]
dirkmc: The idea of having `registerShardAsync` as a separate method is that we want to put the error returned onto the results channel. If we do that inside `registerShardAsync` we have to do the select against context in two places. Doing it this way simplifies the code.

storagemarket/impl/shardregistration.go
raulk: Pushed a solution in https://github.com/filecoin-project/go-fil-markets/commit/f391a57f3af3dd30e79a1052dfcbf1924cb52be6.

Alternatively we could’ve done one iteration where we decided which shards to register (and thus get the size), another to actually register them.

dagstore/dagstorewrapper.go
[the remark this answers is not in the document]
raulk: > The idea of having registerShardAsync as a separate method is that we want to put the error returned onto the results channel.

Why, though? We are anyway waiting for the error to be returned, we could just read it immediately. I'll push a commit for you to review.

raulk (approved):
LGTM for merge into a feature branch for more detailed review along with the rest of the solution. One thing we'll want to do is add info logs to show positive progress during the migration. And we'll want to lower the Errorf here to Warnf because they do not impede progress.

@dirkmc are the test failures just general flakiness?

dirkmc:
Looks like it was just a flaky test, passing now

PR #572, opened 2021-07-14
Integrate CARv2 Blockstore resumption 

aarshkshah1992:
@dirkmc Couple of tests are failing as discussed offline.


dirkmc:
Both these tests are now failing with the same error: `cannot resume from a finalized file`

This is because the CAR file has been finalized but we are trying to open it as a file that can be written to.

aarshkshah1992:
@dirkmc Let's wait till the tests are green.

PR #576, opened 2021-07-18
migrate to DAG store + CARv2 blockstores for storage and retrieval

raulk:
This PR introduces the go-fil-markets side of changes to migrate from a monolithic Badger-backed blockstore, to the [dagstore](https://github.com/filecoin-project/dagstore/), backed by indexed CARv2 files.

This contribution introduces changes in state machines, abstractions, and provider/client environments to remove the reliance on a monolithic blockstore, and instead migrate to a solution backed by the dagstore.

Blockstores are not managed by this module, and are managed by the caller. When this module needs a blockstore in the context of a storage deal or a retrieval, it calls the relevant BlockstoreAccessor.

We believe the test coverage to be pretty satisfactory, with a
substantial portion of the LoC diff corresponding to tests.

This commit is a squashed version of 89 commits developed by @aarshkshah1992, @dirkmc, and @raulk over the course of months.

This contribution was thoroughly tested in the M1 milestone with minerX.2: filecoin-project/lotus#6852.

Read more about the motivation here:
- https://github.com/protocol/web3-dev-team/pull/116
- https://github.com/filecoin-project/dagstore/blob/master/docs/design.

Co-authored-by: @aarshkshah1992 @dirkmc @raulk.

For more information refer to the description in https://github.com/filecoin-project/lotus/pull/6671.


dagstore/dagstorewrapper.go
```diff
@@ -0,0 +1,219 @@
...
+		if xerrors.Unwrap(err) != dagstore.ErrShardUnknown {
+			return nil, xerrors.Errorf("failed to schedule acquire shard for piece CID %s: %w", pieceCid, err)
+		}
+
+		// if the DAGStore does not know about the Shard -> register it and then try to acquire it again.
+		log.Infow("ErrShardUnknown during LoadShard, will re-register", "pieceCID", pieceCid)
+		if err := RegisterShardSync(ctx, ds, pieceCid, "", false); err != nil {
+			return nil, xerrors.Errorf("failed to re-register shard during loading piece CID %s: %w", pieceCid, err)
+		}
+		log.Infow("Successfully re-registered Shard in LoadShard", "pieceCID", pieceCid)
```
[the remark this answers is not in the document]
raulk: nit: Also, remove code references from logging lines. Log lines are intended for operators, and operators don't usually read the code. Prefer:

```
log.Infow("successfully re-registered shard", "pieceCID", pieceCid)
```

In fact, the logging output already prints out the log site as director/file:line.

retrievalmarket/impl/provider_environments.go
```diff
@@ -134,46 +125,17 @@ func (pde *providerDealEnvironment) Node() retrievalmarket.RetrievalProviderNode
...
-	// If there was an error closing the reader, throw that error
-	if closeErr != nil {
-		err := xerrors.Errorf("failed to read file into blockstore: failed to close reader: %w", closeErr)
-		log.Error(err.Error())
-		return err
+// PrepareBlockstore is called when the deal data has been unsealed and we need
+// to add all blocks to a blockstore that is used to serve retrieval
+func (pde *providerDealEnvironment) PrepareBlockstore(ctx context.Context, dealID retrievalmarket.DealID, pieceCid cid.Cid) error {
+	// Load the blockstore that has the deal data
+	bs, err := pde.p.dagStore.LoadShard(ctx, pieceCid)
```
raulk, line 132: Checking: does this execute in its own goroutine, because it might block for a long time while the shard is being fetched and indexed?

raulk (commented):
5-10% in.

raulk (commented):
This is quite a large contribution, so it will take a few passes to review fully ❤️  This is a high-level/architectural pass.

1. IMO, go-fil-markets shouldn't own the DAG store nor the CAR file trackers. go-fil-markets is supposed to be lightweight and reusable as a library, and I would expect no direct side effects on disk.
    - Suggestion: have the DAG store _injected_ into go-fil-markets from Lotus. This also makes parameter configuration (directories, throttling, etc.) easier, as well as future CLI/JSON-RPC operations.
2. Similarly, I'm not convinced that the Lotus mount implementation should live here. go-fil-markets is not Lotus-specific, it's built to be used by other implementations (e.g. Venus), and even independently of Filecoin full node implementations, as a lightweight import (e.g. deal making robots, Estuary, etc.)
    - Suggestion: can we move the Lotus mount to the lotus repo?
3. go-fil-markets shouldn't be doing any file handling or management _itself_. go-fil-markets is intended to be used a library, and this complicates its intergration into third-party software.
    - Suggestion: I think we need an abstraction functionally similar to the Multistore here. I can take a pass and see what I come up with.
4. I'm worried about the profileration of CAR file trackers with apparent code duplication. Can we consolidate the read-only and read-write trackers into a single one?
5. The CAR trackers won't remember state between restarts. Is this intended?
6. The CAR trackers seem to simply abstract over maps. I wonder if they are truly necessary? I'm not sure they provide any value.
7. The CarFileStore also seems a little bit unnecessary. I do see the case for a component that acts like an "on-disk CAR depot". I wonder if it make sense to merge the read-only tracker, read-write tracker, and the CarFileStore into such a component, taking the place of the Multistore?

aarshkshah1992:
> This is quite a large contribution, so it will take a few passes to review fully ❤️ This is a high-level/architectural pass.
> 
> 1. IMO, go-fil-markets shouldn't own the DAG store nor the CAR file trackers. go-fil-markets is supposed to be lightweight and reusable as a library, and I would expect no direct side effects on disk.
>    

>    * Suggestion: have the DAG store _injected_ into go-fil-markets from Lotus. This also makes parameter configuration (directories, throttling, etc.) easier, as well as future CLI/JSON-RPC operations.

I agree with this.

> 2. Similarly, I'm not convinced that the Lotus mount implementation should live here. go-fil-markets is not Lotus-specific, it's built to be used by other implementations (e.g. Venus), and even independently of Filecoin full node implementations, as a lightweight import (e.g. deal making robots, Estuary, etc.)
>    
>    * Suggestion: can we move the Lotus mount to the lotus repo?

I agree with this.

> 3. go-fil-markets shouldn't be doing any file handling or management _itself_. go-fil-markets is intended to be used a library, and this complicates its intergration into third-party software.
>    
>    * Suggestion: I think we need an abstraction functionally similar to the Multistore here. I can take a pass and see what I come up with.

We do have the `filestore` that Markets can use for file handling/management. I think we can re-use it for all the file handling in here. If we're not, we're doing it wrong.

> 4. I'm worried about the profileration of CAR file trackers with apparent code duplication. Can we consolidate the read-only and read-write trackers into a single one?

I can give it a shot and see what I can come up with.

> 5. The CAR trackers won't remember state between restarts. Is this intended?

Yes, this is in-memory. The CAR file path is persisted to disk as part of the deal state and we can always recreate/resume the CAR blockstore from CAR file.

> 6. The CAR trackers seem to simply abstract over maps. I wonder if they are truly necessary? I'm not sure they provide any value.

 It's nice to have. If we get rid of these trackers, we'd have to put this (deal -> CAR blockstore) creation, state management and the corresponding blockstore  cleanup  codein the  Storage Client, Storage Provider, Retrieval Provider and Retrieval Client. We don't wanna copy the same code in all the four places.


> 7. The CarFileStore also seems a little bit unnecessary. I do see the case for a component that acts like an "on-disk CAR depot". I wonder if it make sense to merge the read-only tracker, read-write tracker, and the CarFileStore into such a component, taking the place of the Multistore?

This would be nice to have for sure. I am just worried about building a new component, testing it, integrating it and then testing the whole thing again in a span of 3 days while we are also focused on M1 testing. But, we can discuss this.

dagstore/dagstorewrapper.go
```diff
@@ -0,0 +1,219 @@
...
+	"github.com/filecoin-project/dagstore/shard"
+
+	"github.com/filecoin-project/go-fil-markets/carstore"
+)
+
+var log = logging.Logger("dagStoreWrapper")
+var gcInterval = 5 * time.Minute
+
+// MarketDAGStoreConfig is the config the market needs to then construct a DAG Store.
+type MarketDAGStoreConfig struct {
```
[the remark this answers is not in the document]
aarshkshah1992: I agree. We should do this.

dagstore/dagstorewrapper.go
```diff
@@ -0,0 +1,219 @@
...
+
+// DagStore provides an interface to the concrete DAG store that can be mocked
+// out by tests
+type DagStore interface {
+	RegisterShard(ctx context.Context, key shard.Key, mnt mount.Mount, out chan dagstore.ShardResult, opts dagstore.RegisterOpts) error
+	AcquireShard(ctx context.Context, key shard.Key, out chan dagstore.ShardResult, _ dagstore.AcquireOpts) error
+	RecoverShard(ctx context.Context, key shard.Key, out chan dagstore.ShardResult, _ dagstore.RecoverOpts) error
+	GC(ctx context.Context) (map[shard.Key]error, error)
+	Close() error
+}
```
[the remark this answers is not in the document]
aarshkshah1992: @raulk That would be great. I suspect most DAGStore clients would want this.

carstore/read_write_blockstore.go
```diff
@@ -0,0 +1,66 @@
...
+	"sync"
+
+	"github.com/ipfs/go-cid"
+	"github.com/ipld/go-car/v2/blockstore"
+	"golang.org/x/xerrors"
+)
+
+// CarReadWriteStoreTracker tracks the lifecycle of a ReadWrite CAR Blockstore and makes it easy to create/get/cleanup the blockstores.
+// It's important to close a CAR Blockstore when done using it so that the backing CAR file can be closed.
+type CarReadWriteStoreTracker struct {
```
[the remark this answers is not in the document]
aarshkshah1992: **_Note:_** Retrieval resumption is NOT supported for process restarts even before this upgrade.

- For connection restarts, things will work fine.
- Having said that, given that the CARv2 read-write blockstore supports resumption from a partially written CARv2 file, we can use this as is when we do build out the "resumption from a process bounce" functionality for the Retrieval Markets.

dagstore/dagstorewrapper.go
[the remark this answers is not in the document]
aarshkshah1992: @raulk This can happen if the upfront shard migration on the Storage Provider after the upgrade fails for whatever reason. I remember we'd discussed some rare but hairy edge cases there that can cause this.

I agree with your recommendations here and am incorporating them in the the upcoming PR.

retrievalmarket/impl/provider_environments.go
aarshkshah1992: Yes, this executes in the "per-deal" go-routine that the Markets fsm spins up and will NOT block other deals.

filestorecaradapter/adapter.go
```diff
@@ -0,0 +1,87 @@
...
+// NewReadOnlyFileStore returns a Filestore backed by the given CAR file that clients can read Unixfs DAG blocks from.
+// The intermediate blocks of the Unixfs DAG are returned as is from the given CAR file.
+// For the leaf nodes of the Unixfs DAG, if  `PosInfo`(filepath, offset, size) nodes have been written to the CAR file,
+// the filestore will read the `PosInfo` node from the CAR file and then resolve the actual raw leaf data from the file
+// referenced by the `PosInfo` node.
+//
+// Note that if the given CAR file does NOT contain any `PosInfo` nodes and contains all Unixfs DAG blocks
+// as is, the filestore will return all blocks as is from the given CAR file i.e. in such a case,
+// the Filestore will simply act as a pass-through read only CAR blockstore.
+func NewReadOnlyFileStore(carFilePath string) (*FileStore, error) {
```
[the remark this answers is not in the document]
aarshkshah1992: @raulk The thing is, we can give this a normal/full CAR file that is NOT linked to a regular file and things will still work fine for clients such as the UnixFS DAG library etc.  We also want clients to be aware that they're using a Filestore here. 

Do you still wanna go ahead with renaming this to a `LinkedCAR`/have a better name in mind ?

filestorecaradapter/internal/adapter.go
```diff
@@ -0,0 +1,89 @@
...
+	panic("implement me")
+}
+
+func (a *bsToDSBatchingAdapter) Sync(prefix datastore.Key) error {
+	panic("implement me")
+}
+
+func (a *bsToDSBatchingAdapter) Close() error {
+	return nil
+}
```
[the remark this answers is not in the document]
aarshkshah1992: Agreed.

filestore/carfilestore.go
```diff
@@ -0,0 +1,23 @@
+package filestore
```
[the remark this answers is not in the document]
aarshkshah1992: @dirkmc Could you elicit why we need to introduce this additional abstraction and not use the `filestore` that we already have here ?

dagstore/dagstorewrapper.go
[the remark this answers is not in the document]
aarshkshah1992: @dirkmc is working on this.

filestore/carfilestore.go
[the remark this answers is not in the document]
dirkmc: The CarFileStore was introduced to provide a way to inject a mechanism to map from a deal ID to a file path.
In practice all we need is a directory path that we can append to, this is a case of YAGNI :)
I replace it with a directory path in commit https://github.com/filecoin-project/go-fil-markets/pull/576/commits/11b5719077d99053675735b8ec4ca5acf892f8a8

go.mod
```diff
@@ -3,24 +3,30 @@ module github.com/filecoin-project/go-fil-markets
...
+	github.com/filecoin-project/specs-actors/v2 v2.3.5-0.20210114162132-5b58b773f4fb
+	github.com/filecoin-project/specs-storage v0.1.1-0.20201105051918-5188d9774506
+	github.com/golang/mock v1.4.4
 	github.com/hannahhoward/cbor-gen-for v0.0.0-20200817222906-ea96cece81f1
 	github.com/hannahhoward/go-pubsub v0.0.0-20200423002714-8d62886cc36e
-	github.com/ipfs/go-block-format v0.0.2
-	github.com/ipfs/go-blockservice v0.1.4-0.20200624145336-a978cec6e834
-	github.com/ipfs/go-cid v0.0.7
+	github.com/ipfs/go-block-format v0.0.3
+	github.com/ipfs/go-blockservice v0.1.4
```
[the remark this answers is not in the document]
aarshkshah1992: Fixed in https://github.com/filecoin-project/go-fil-markets/pull/575.

go.mod
```diff
@@ -3,24 +3,30 @@ module github.com/filecoin-project/go-fil-markets
...
 	github.com/filecoin-project/go-data-transfer v1.7.0
 	github.com/filecoin-project/go-ds-versioning v0.1.0
 	github.com/filecoin-project/go-multistore v0.0.3
 	github.com/filecoin-project/go-padreader v0.0.0-20200903213702-ed5fae088b20
-	github.com/filecoin-project/go-state-types v0.0.0-20201102161440-c8033295a1fc
+	github.com/filecoin-project/go-state-types v0.1.1-0.20210506134452-99b279731c48
 	github.com/filecoin-project/go-statemachine v0.0.0-20200925024713-05bd7c71fbfe
-	github.com/filecoin-project/go-statestore v0.1.0
+	github.com/filecoin-project/go-statestore v0.1.1
+	github.com/filecoin-project/lotus v1.10.1
```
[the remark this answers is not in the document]
aarshkshah1992: This was because of how we were checking the Unsealing status earlier in the Shard migration code. Has been fixed in https://github.com/filecoin-project/go-fil-markets/pull/575.

dagstore/dagstorewrapper.go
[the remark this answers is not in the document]
aarshkshah1992: Fixed in https://github.com/filecoin-project/go-fil-markets/pull/577.

filestorecaradapter/internal/adapter.go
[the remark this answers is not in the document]
aarshkshah1992: Fixed in https://github.com/filecoin-project/go-fil-markets/pull/577.

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/filecoin-project/go-fil-markets
 go 1.13
 
 require (
-	github.com/filecoin-project/dagstore v0.2.0
+	github.com/filecoin-project/dagstore v0.2.1-0.20210720172151-1ebca744487c
```
masih, line 6: Should this depend on `v2.0.0-beta1` tag?

carstore/read_only_blockstore.go
```diff
@@ -48,7 +48,7 @@ func (r *CarReadOnlyStoreTracker) GetOrCreate(key string, carFilePath string) (C
 		return bs, nil
 	}
 
-	rdOnly, err := blockstore.OpenReadOnly(carFilePath, carv2.ZeroLengthSectionAsEOF)
+	rdOnly, err := blockstore.OpenReadOnly(carFilePath, carv2.ZeroLengthSectionAsEOF(true))
```
masih, line 51: Do we need to set option `blockstore.UseWholeCIDs(true)` here?

Ditto in other places that instantiate CARv2 blockstore.

go.mod
aarshkshah1992: Have updated this.

carstore/read_only_blockstore.go
aarshkshah1992: Have done this here and at all the other places. Thanks !

filestorecaradapter/adapter.go
```diff
@@ -0,0 +1,88 @@
+package filestorecaradapter
```
raulk, line 1: This package should probably move to the go-filestore library.

filestorecaradapter/adapter.go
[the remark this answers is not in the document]
raulk: I'm not sure I like this being so general. LinkedCAR is only used in one particular setting, and that's the client import. Right? Not sure any of that logic should be so deep in this library (go-fil-markets). This lib should not handle file paths at all.

storagemarket/impl/provider_environments.go
```diff
@@ -32,6 +33,27 @@ type providerDealEnvironment struct {
...
+func (p *providerDealEnvironment) RegisterShard(ctx context.Context, pieceCid cid.Cid, carPath string, eagerInit bool) error {
+	return shared.RegisterShardSync(ctx, p.p.dagStore, pieceCid, carPath, eagerInit)
+}
+
+func (p *providerDealEnvironment) CARv2Reader(carV2FilePath string) (*carv2.Reader, error) {
+	return carv2.OpenReader(carV2FilePath)
+}
+
+func (p *providerDealEnvironment) FinalizeReadWriteBlockstore(proposalCid cid.Cid, carPath string, rootCid cid.Cid) error {
+	bs, err := p.p.readWriteBlockStores.GetOrCreate(proposalCid.String(), carPath, rootCid)
```
raulk, line 45: Shouldn't this be a Get?

raulk: If so, the signature of this method can take only the proposalCid.

storagemarket/impl/provider_environments.go
```diff
@@ -48,48 +70,73 @@ func (p *providerDealEnvironment) Ask() storagemarket.StorageAsk {
 	return *sask.Ask
 }
 
-func (p *providerDealEnvironment) DeleteStore(storeID multistore.StoreID) error {
-	return p.p.multiStore.Delete(storeID)
+func (p *providerDealEnvironment) CleanReadWriteBlockstore(proposalCid cid.Cid, carV2FilePath string) error {
```
raulk, line 73: We should not need to pass the path here. We should store the path in the tracker itself.

carstore/read_only_blockstore.go
```diff
@@ -0,0 +1,83 @@
...
+
+	if _, ok := r.stores[key]; ok {
+		return false, nil
+	}
+
+	r.stores[key] = bs
+	return true, nil
+}
+
+func (r *CarReadOnlyStoreTracker) GetOrCreate(key string, carFilePath string) (ClosableBlockstore, error) {
```
raulk, line 43: Is there even a valid use case for this method? This will always fail.

raulk: I misunderstood. This should probably be called GetOrOpen.

carstore/read_only_blockstore.go
```diff
@@ -0,0 +1,83 @@
...
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	if bs, ok := r.stores[key]; ok {
+		return bs, nil
+	}
+
+	rdOnly, err := blockstore.OpenReadOnly(carFilePath, carv2.ZeroLengthSectionAsEOF(true), blockstore.UseWholeCIDs(true))
```
raulk, line 51: Should this be wrapped in a Filestore?

storagemarket/impl/provider_environments.go
raulk: Alright, so this can't be a Get because it's possible that we restarted, and we are not tracking the store any longer. Ideally we would track stores for all pending transfers.

carstore/read_only_blockstore.go
```diff
@@ -0,0 +1,83 @@
...
+	defer r.mu.RUnlock()
+
+	if bs, ok := r.stores[key]; ok {
+		return bs, nil
+	}
+
+	return nil, xerrors.Errorf("could not get blockstore for key %s: %w", key, ErrNotFound)
+}
+
+func (r *CarReadOnlyStoreTracker) CleanBlockstore(key string) error {
```
raulk, line 71: This is a bit of a misnomer because it's not "cleaning" anything; it's just untracking it.

raulk: Will rename.

retrievalmarket/impl/integration_test.go
```diff
@@ -275,7 +280,7 @@ func TestClientCanMakeDealWithProvider(t *testing.T) {
 		{name: "multi-block file retrieval succeeds",
 			filename:    "lorem.txt",
 			filesize:    19000,
-			voucherAmts: []abi.TokenAmount{abi.NewTokenAmount(10136000), abi.NewTokenAmount(19920000)},
+			voucherAmts: []abi.TokenAmount{abi.NewTokenAmount(10174000), abi.NewTokenAmount(19958000)},
```
hannahhoward, line 283: I must be dense but I've read through this multiple times and still cant figure out why the vouchers changed. It appears (I think) that everything went up by exactly 38000? Which happens to be 2x the filesize? I'd like to understand why.

retrievalmarket/impl/integration_test.go
```diff
@@ -385,19 +412,12 @@ func TestClientCanMakeDealWithProvider(t *testing.T) {
...
-			require.NoError(t, err)
-			carData := buf.Bytes()
+			// Set up the piece info that will be retrieved by the provider
+			// when the retrieval request is made
 			sectorID := abi.SectorNumber(100000)
 			offset := abi.PaddedPieceSize(1000)
-			pieceInfo = piecestore.PieceInfo{
-				PieceCID: tut.GenerateCids(1)[0],
+			pieceInfo := piecestore.PieceInfo{
+				PieceCID: payloadCID,
```
hannahhoward, line 420: Why is it neccesary to change this? Has the use of PieceInfo changed in some manner? If not, it's super confusing to use payloadCID here -- cause it's not the same thing -- the switch makes me think this means something, but I don't know.

retrievalmarket/storage_retrieval_integration_test.go
```diff
@@ -61,7 +63,7 @@ func TestStorageRetrieval(t *testing.T) {
 			pricePerByte:            abi.NewTokenAmount(1000),
 			paymentInterval:         uint64(10000),
 			paymentIntervalIncrease: uint64(1000),
-			voucherAmts:             []abi.TokenAmount{abi.NewTokenAmount(10136000), abi.NewTokenAmount(19920000)},
+			voucherAmts:             []abi.TokenAmount{abi.NewTokenAmount(10174000), abi.NewTokenAmount(19958000)},
```
hannahhoward, line 66: still confused about the voucher changes

shared_testutil/mockdagstorewrapper.go
```diff
@@ -0,0 +1,136 @@
...
+
+	// Get the piece info from the piece store
+	pi, err := m.pieceStore.GetPieceInfo(pieceCid)
+	if err != nil {
+		return nil, err
+	}
+
+	// Unseal the sector data for the deal
+	deal := pi.Deals[0]
+	r, err := m.rpn.UnsealSector(ctx, deal.SectorID, deal.Offset.Unpadded(), deal.Length.Unpadded())
```
hannahhoward, line 105: Two things:
1. Why go through all this chicanery for something that's a mock implementation to begin with?
2. My guess from reading the code is that the dagstore hides the Unsealing behind LoadShard. (or some wrapper in Lotus does) and as far as I can tell UnsealSector, the method on the retrieval provider node, is no longer used, except oddly in this test fixture. If this is the case, I would say, remove it. We should not have methods on the node that aren't used, as it means a node implementer must support them.

hannahhoward (commented):
Mid review notes (through to storage market):

Almost all my comments are non blocking, however:
- I'd like to get the stores/filestore_test duplicate logic either fixed or understand why it's not duplicate
- I'd like to get answers to: 
   - why all the test voucher amounts changed
   - whether it means anything that the variable used for PieceCID under piece info in the integration test changed
 
 Stuff that I'm fine merging, but we should really consider seperate future cleanup tickets for:
 - finding a simply solution to the lazy blockstore
 - dropping the UnsealSector method if we don't use it
 
 Will continue with more notes when I'm done.

storagemarket/impl/provider_environments.go
```diff
@@ -48,48 +85,58 @@ func (p *providerDealEnvironment) Ask() storagemarket.StorageAsk {
...
+				return
+			}
+		}
+	}()
 
-	if p.p.universalRetrievalEnabled {
-		pieceCid, psize, path, err = providerutils.GeneratePieceCommitmentWithMetadata(p.p.fs, p.p.pio.GeneratePieceCommitment, proofType, payloadCid, selector, storeID)
-	} else {
-		pieceCid, psize, err = p.p.pio.GeneratePieceCommitment(proofType, payloadCid, selector, storeID)
+	// dump the CARv1 payload of the CARv2 file to the Commp Writer and get back the CommP.
```
hannahhoward, line 110: If `rd.DataReader()` is an `io.Reader`,  the `io.Copy` can be avoided by just using padreader and the FFI directly. Not blocking for merge but this seems like and optimization worth pursuing.

All the complicated machinery from before is due to having to write the CAR file from the blockstore and then read it to compute CommP. Here we already have a reader.

Cribbed from source code inside writer --

``` golang
import (
	ffi "github.com/filecoin-project/filecoin-ffi"
	"github.com/filecoin-project/go-padreader"
)

r, sz := padreader.New(rd.DataReader(), rd.Header.DataSize)
pieceCid, err := ffiwrapper.GeneratePieceCIDFromFile(abi.RegisteredSealProof_StackedDrg32GiBV1, r, sz)
pieceSize := sz.Padded()
```

storagemarket/impl/clientutils/clientutils.go
```diff
@@ -5,37 +5,69 @@ import (
...
 	}
 
-	commp, paddedSize, err := pieceIO.GeneratePieceCommitment(rt, data.Root, shared.AllSelector(), storeID)
+	if carPath == "" {
+		return cid.Undef, 0, xerrors.New("need Carv2 file path to get a read-only blockstore")
+	}
+
+	// Open a read-only blockstore off the CAR file, wrapped in a filestore so
+	// it can read file positional references.
+	fs, err := stores.ReadOnlyFilestore(carPath)
```
hannahhoward, line 46: I put a more in depth comment on the provider side, but I suspect there is an optimization here as well to avoid reading from a car then writing that car then reading to calculate commP. I think the optimization can be deferred but it is worth pursuing.

storagemarket/impl/provider_environments_test.go
```diff
@@ -0,0 +1,50 @@
...
+	_, carV2File2 := shared_testutil.CreateDenseCARv2(t, filepath.Join("storagemarket", "fixtures", "payload2.txt"))
+	defer os.Remove(carV2File2)
+
+	commP1 := genProviderCommP(t, carV2File1, pieceSize)
+	commP2 := genProviderCommP(t, carV2File2, pieceSize)
+
+	commP3 := genProviderCommP(t, carV2File1, pieceSize)
+	commP4 := genProviderCommP(t, carV2File2, pieceSize)
+
+	require.Equal(t, commP1, commP3)
```
hannahhoward, line 30: I do not understand the value of this test -- are we surprised that to runs of the same function on the same file generate the same output while two runs different files produce different output? That seems somewhat self-evident. If the goal is to test that is actually produces "the right output" I feel like this test is incomplete. However, I also feed the loginc in the code itself is small enough that I'm not sure why we need to test this, especially since it primarily relies on code from commp-utils

storagemarket/impl/providerstates/provider_states_test.go
```diff
@@ -1122,6 +1129,35 @@ var defaultMetadataFile = tut.NewTestFile(tut.TestFileParams{
 	Size:   400,
 })
 
+func mkPieceCid(input string) cid.Cid {
```
hannahhoward, line 1132: Non-blocking but I would say the proper solution is to move common fixture generation out of specs-actors so lots of libraries can use them. For another day.

storagemarket/impl/providerstates/provider_fsm.go
```diff
@@ -261,3 +261,13 @@ var ProviderFinalityStates = []fsm.StateKey{
 	storagemarket.StorageDealSlashed,
 	storagemarket.StorageDealExpired,
 }
+
+// StatesKnownBySealingSubsystem are the states on the happy path after hand-off to
+// the sealing subsystem
+var StatesKnownBySealingSubsystem = []fsm.StateKey{
```
hannahhoward, line 267: What is this used for? I can't find it anywhere in the code? I feel like if it's used in the markets module, having a public var a deeply nested section is maybe not the best idea?

hannahhoward (changes requested):
Review at end:

NO additional blocking changes but I would like to get some answers to above questions about the retrieval side and resolve my understanding re: the duplicate tests in the one retrieval file.

I'd also like to see some of my suggestions both on retrieval and now on storage for post-merge cleanup acted upon, but I do not want to hold up merge and release any further.

storagemarket/impl/client.go
```diff
@@ -78,26 +75,21 @@ func DealPollingInterval(t time.Duration) StorageClientOption {
...
-	bs blockstore.Blockstore,
-	multiStore *multistore.MultiStore,
 	dataTransfer datatransfer.Manager,
 	discovery *discoveryimpl.Local,
 	ds datastore.Batching,
 	scn storagemarket.StorageClientNode,
 	options ...StorageClientOption,
 ) (*Client, error) {
-	carIO := cario.NewCarIO()
-	pio := pieceio.NewPieceIO(carIO, bs, multiStore)
```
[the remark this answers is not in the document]
dirkmc: Yes I think we can delete it from go-commp-utils 👍

storagemarket/impl/providerstates/provider_fsm.go
dirkmc: It's used when registering deals with the DAG store - we only want to register deals that have been handed off to the sealing subsystem (not deals that haven't yet reached that stage, or that have reached a terminal state).
I think it makes sense to have it close to the FSM, so that if more states are added to the FSM it's more likely to be noticed that they need to be added to this array as well.

retrievalmarket/impl/client.go
```diff
@@ -72,16 +75,17 @@ func dispatcher(evt pubsub.Event, subscriberFn pubsub.SubscriberFn) error {
 var _ retrievalmarket.RetrievalClient = &Client{}
 
 // NewClient creates a new retrieval client
-func NewClient(network rmnet.RetrievalMarketNetwork, multiStore *multistore.MultiStore, dataTransfer datatransfer.Manager, node retrievalmarket.RetrievalClientNode, resolver discovery.PeerResolver, ds datastore.Batching) (retrievalmarket.RetrievalClient, error) {
+func NewClient(network rmnet.RetrievalMarketNetwork, carPath string, dataTransfer datatransfer.Manager, node retrievalmarket.RetrievalClientNode, resolver discovery.PeerResolver, ds datastore.Batching) (retrievalmarket.RetrievalClient, error) {
```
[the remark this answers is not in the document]
dirkmc: The retrieval client puts files in the directory that carPath points to. It's just a little simpler for the caller if it specifies a directory to put stuff in rather than having to specify a new filename every time.

retrievalmarket/impl/provider_environments.go
```diff
@@ -330,12 +299,25 @@ type providerStoreGetter struct {
...
+	// When a request for data is received
+	// 1. The data transfer layer calls Get to get the blockstore
+	// 2. The data for the deal is unsealed
+	// 3. The unsealed data is put into the blockstore (in this case a CAR file)
+	// 4. The data is served from the blockstore (using blockstore.Get)
+	//
+	// So we use a "lazy" blockstore that can be returned in step 1
+	// but is only accessed in step 4 after the data has been unsealed.
+	//
+	return newLazyBlockstore(func() (dagstore.ReadBlockstore, error) {
```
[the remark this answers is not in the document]
dirkmc: I agree, the lazy blockstore is really a hack around the way that the data transfer layer assumes that the blockstore works (it assumes it works like a multistore). The correct fix would be to change the assumption at the data transfer layer.

retrievalmarket/impl/integration_test.go
dirkmc: Good catch 🎣 

In the test we were incorrectly expecting the payload CID instead of the piece CID. Fixed in https://github.com/filecoin-project/go-fil-markets/pull/576/commits/094138303c8d9a33bd525bde129313938b173221

shared_testutil/mockdagstorewrapper.go
dirkmc: You're right that the dag store now takes care of unsealing internally. So in order to test that the deal flow works end to end, including the unsealing part, we implemented this mock dag store wrapper.

dagstore/dagstorewrapper.go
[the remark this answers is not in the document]
raulk: Done on dagstore side.

retrievalmarket/impl/client.go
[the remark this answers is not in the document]
raulk: I agree with Hannah, and I also pointed out in other comments that this library should not be actively managing files on the filesystem. I too agree with @dirkmc that it's a harder change to implement, but it shouldn't be too hard. It's just good behaviour to let the user specify where they want the output of the operation (CAR file) placed.

stores/filestore.go
```diff
@@ -0,0 +1,161 @@
...
+
+	bs, err := FilestoreOf(ro)
+	if err != nil {
+		return nil, err
+	}
+
+	return &closableBlockstore{Blockstore: bs, closeFn: ro.Close}, nil
+}
+
+// ReadWriteFilestore opens the CAR in the specified path as as a read-write
```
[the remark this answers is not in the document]
raulk: Yep, that's it.

retrievalmarket/impl/integration_test.go
raulk: Good question -- @dirkmc could you investigate this? It does sound like an unexpected functional change.

retrievalmarket/storage_retrieval_integration_test.go
raulk: 👍  @dirkmc let's follow up on this!

storagemarket/impl/client.go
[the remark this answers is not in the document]
raulk: Ok, that works for me. I thought we wanted that piece to be reusable, hence placing it in a separate library. But I'll mark the related open TODO as done. @dirkmc -- are you taking this on?

storagemarket/impl/provider.go
```diff
@@ -590,30 +575,47 @@ func (p *Provider) start(ctx context.Context) error {
 	err := p.migrateDeals(ctx)
 	publishErr := p.readyMgr.FireReady(err)
```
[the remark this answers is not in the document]
raulk: Yes, I think you're right. @dirkmc

storagemarket/impl/provider_environments.go
```diff
@@ -32,6 +33,42 @@ type providerDealEnvironment struct {
 	p *Provider
 }
 
+func (p *providerDealEnvironment) RegisterShard(ctx context.Context, pieceCid cid.Cid, carPath string, eagerInit bool) error {
+	return stores.RegisterShardSync(ctx, p.p.dagStore, pieceCid, carPath, eagerInit)
+}
+
+func (p *providerDealEnvironment) ReadCAR(path string) (*carv2.Reader, error) {
```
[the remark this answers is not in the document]
raulk: Agree, but I wonder if we're reimplementing this in the mock with different logic?

storagemarket/impl/provider_environments.go
raulk: @dirkmc can you follow-up on this?

storagemarket/impl/clientutils/clientutils.go
raulk: @dirkmc follow-up here?

storagemarket/impl/providerstates/provider_fsm.go
raulk: Agree with @dirkmc, we do want to keep it close to the authority in states (go-fil-markets).

storagemarket/impl/provider.go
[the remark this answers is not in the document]
dirkmc: https://github.com/filecoin-project/go-fil-markets/pull/576/commits/77be1e6abdd8b373481f654cd04f1d22fcfc72fc

retrievalmarket/impl/client.go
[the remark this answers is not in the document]
dirkmc: Currently the carPath is initialized as a directory inside the DAG store:
```
carsPath := filepath.Join(r.Path(), DefaultDAGStoreDir, "retrieval-cars")
```

And the file path is derived from the deal ID:
```
carFilePath := filepath.Join(c.carPath, dealID.String())
```

This has the nice property that the name of the file containing the retrieval data is the same as the deal ID.

If we instead pass the file path as a parameter to the `Retrieve()` function we will not be able to name the file the same as the deal ID (the deal ID gets created inside the `Retrieve()` function.

retrievalmarket/impl/integration_test.go
dirkmc: To be honest I'm not sure either. This is the commit where the change was made:
https://github.com/filecoin-project/go-fil-markets/commit/cf7e2a9ecb1cda8ae5331c8a248b3a7f3d54ac97
But I'm not sure why the amounts are different.

storagemarket/impl/provider_environments.go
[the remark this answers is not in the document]
dirkmc: Looks like we are mocking this out so we can test the error case

storagemarket/impl/provider_environments_test.go
dirkmc: Good question!
Will follow up with @aarshkshah1992

retrievalmarket/impl/client.go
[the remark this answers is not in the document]
dirkmc: Discussed sync with @raulk 
Resolution is to adopt Hannah's suggestion

stores/filestore_test.go
```diff
@@ -0,0 +1,206 @@
...
+	require.EqualValues(t, origBytes, fbz)
+}
+
+func TestReadOnlyFilestoreWithDenseCARFile(t *testing.T) {
+	ctx := context.Background()
+	normalFilePath, origContent := createFile(t, 10, 10485760)
+
+	// write out a unixfs dag to an inmemory store to get the root.
+	root := writeUnixfsDAGInmemory(t, ctx, normalFilePath)
+	require.NotEqualValues(t, cid.Undef, root)
```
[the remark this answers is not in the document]
dirkmc: @raulk please take a look 👁️👁️

stores/filestore_test.go
```diff
@@ -0,0 +1,206 @@
...
+
+var defaultHashFunction = uint64(mh.BLAKE2B_MIN + 31)
+
+func TestReadOnlyFilstoreWithPosInfoCARFile(t *testing.T) {
+	ctx := context.Background()
+	normalFilePath, origBytes := createFile(t, 10, 10485760)
+
+	// write out a unixfs dag to an inmemory store to get the root.
+	root := writeUnixfsDAGInmemory(t, ctx, normalFilePath)
+	require.NotEqualValues(t, cid.Undef, root)
```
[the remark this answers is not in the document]
dirkmc: @raulk please take a look 👁️👁️

stores/filestore_test.go
```diff
@@ -0,0 +1,206 @@
...
+)
+
+const (
+	unixfsChunkSize     uint64 = 1 << 10
+	unixfsLinksPerLevel        = 1024
+)
+
+var defaultHashFunction = uint64(mh.BLAKE2B_MIN + 31)
+
+func TestReadOnlyFilstoreWithPosInfoCARFile(t *testing.T) {
```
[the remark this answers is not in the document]
dirkmc: @raulk please take a look 👁️👁️

retrievalmarket/impl/provider_environments.go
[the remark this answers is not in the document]
dirkmc: Resolution: it's hacky but beyond the scope of this PR to rework how go-data-transfer manages blockstores so let's go with the lazy blockstore solution for now.

retrievalmarket/impl/integration_test.go
dirkmc: @raulk please take a look 👁️👁️

storagemarket/impl/client.go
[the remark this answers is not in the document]
dirkmc: Resolution: Dirk to remove unused code from go-commp-utils

storagemarket/impl/provider_environments.go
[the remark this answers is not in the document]
dirkmc: We are testing the error case using a mock

storagemarket/impl/provider_environments.go
dirkmc: @raulk please take a look 👁️👁️

storagemarket/impl/clientutils/clientutils.go
dirkmc: @raulk please take a look 👁️👁️

retrievalmarket/impl/client.go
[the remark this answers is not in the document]
hannahhoward: follow-up -- upon further review of the Lotus code, should Retrieve just take a blockstore? This might make IPFS blockstore integration easier.

shared_testutil/mockdagstorewrapper.go
hannahhoward: I would say: a mock implementation is a mock implementation, and having it do fake implementations that themselves call mock functions (UnsealSector is a mock implementation) doesn't actually add value to testing code.

The assumption is that the boundary of testing here is what's in the go-fil-markets library, that the dagstore is unit tested, and the rest is handled by lotus integration testing.

storagemarket/nodes.go
```diff
@@ -105,6 +105,10 @@ type StorageProviderNode interface {
 
 	// GetProofType gets the current seal proof type for the given miner.
 	GetProofType(ctx context.Context, addr address.Address, tok shared.TipSetToken) (abi.RegisteredSealProof, error)
+
+	// IsUnsealed checks if we have an unsealed sector file for the given sector which contains the unsealed deal data
```
hannahhoward, line 109: Follow-up: Why was this added? It appears to be unused in this code.

hannahhoward (commented):
See follow-up comment about IsUnsealed after reviewing Lotus PR.

retrievalmarket/impl/client.go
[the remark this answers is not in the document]
dirkmc: It needs to be a file so that we can safely support resumption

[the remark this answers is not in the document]
dirkmc: Fixed in https://github.com/filecoin-project/go-fil-markets/pull/576/commits/6b375f9a43cf6c233c5722310c296e330126a861

retrievalmarket/impl/lazyblockstore.go
```diff
@@ -0,0 +1,94 @@
...
+func (l *lazyBlockstore) HashOnRead(enabled bool) {
+	bs, err := l.init()
+	if err != nil {
+		return
+	}
+	bs.HashOnRead(enabled)
+}
+
+func (l *lazyBlockstore) init() (dagstore.ReadBlockstore, error) {
+	l.lk.Lock()
```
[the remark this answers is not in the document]
dirkmc: I don't think there's an easy way to also capture and store the error with sync.Once()

storagemarket/nodes.go
dirkmc: Good catch 👍 

Removed in https://github.com/filecoin-project/go-fil-markets/pull/576/commits/c6cd9ae4259fb3eb477dfd7f091f7c8432aec63a

stores/filestore_test.go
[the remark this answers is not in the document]
raulk: My bad, this was my fault when I was refactoring. Fixed in 4a308b5. Thanks @hannahhoward!

stores/filestore_test.go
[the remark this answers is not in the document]
raulk: But I'll remove anyway.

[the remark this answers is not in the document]
raulk: Fixed in 901c837.

shared_testutil/mockdagstorewrapper.go
raulk: However, I think the reason why this is needed is because even "inside" the test framework, the component that custodies the pieces is the `PieceStore`, and there's a mock for that. Same for the `SectorAccessor`. And a test that uses this test framework would reasonably assume that it can reuse those bits to mock out the piece and sector layers, and have those feed into the rest of the test framework. Unfortunately that leads to a MockDagStoreWrapper that actually holds a little bit of logic (which is not illegal IMO, because it's just satisfying the contract of the interface in the advertised manner), but I can't think of another way.

storagemarket/impl/client.go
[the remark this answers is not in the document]
raulk: This can be done later.

storagemarket/impl/provider_environments.go
raulk: Two things here, @hannahhoward:

1. Are you suggesting to stop using `go-commp-utils/writer.Writer` in favour of this more direct approach? I see that `writer.Writer` implements some parallelisation. Does the recipe you suggest also do that? Sorry but I'm not familiar with the motivation and the existence of these utilities, and unfortunately they are not documented either.
2. Are you sure you want a direct `filecoin-ffi` dependency from `go-fil-markets`? Right now it's not on the direct build graph.

storagemarket/impl/providerstates/provider_states_test.go
raulk: Yep, not right now.

stores/filestore_test.go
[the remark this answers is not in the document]
raulk: Removed.

retrievalmarket/impl/integration_test.go
raulk: Following investigation, this change is correct. We modified the `DagBuilderParams` to match those used in `lotus client import` for a more realistic test, and that leads to a heavier DAG, thus higher cost.

If I revert that change, the voucher amounts revert to the previous ones.

Conclusion: side-effect of the test changing (for the better).

image

https://github.com/filecoin-project/go-fil-markets/commit/cf7e2a9ecb1cda8ae5331c8a248b3a7f3d54ac97#diff-26f3d1665dab327881c8abd11dfa4b1e249f730bddfefdec7363fe08290148fdR113-R122

retrievalmarket/storage_retrieval_integration_test.go
raulk: See above. Nothing fishy here.

hannahhoward (approved):
Pre-emptive approval to unblock merge while I'm on vacation. I will double check any changes in tomorrow.

storagemarket/impl/provider_environments.go
raulk: Confirmed offline with @hannahhoward that we don't want to stop using `writer.Writer`.

retrievalmarket/impl/integration_test.go
aarshkshah1992: @raulk is correct here.

PR #596, opened 2021-08-03
review of dagstore + carv2

raulk:
- consolidation of small packages related to blockstore management and filestores under package `stores`.
- naming changes for more succinctness.
- simplify blockstore trackers; revise naming.
- storage provider: register ongoing CAR blockstores on restart; more obvious than the previous behaviour.

## TODO

Refactoring migration.


storagemarket/impl/client.go
```diff
@@ -374,16 +374,16 @@ func (c *Client) ProposeStorageDeal(ctx context.Context, params storagemarket.Pr
...
+		ProposalCid:        proposalNd.Cid(),
+		ClientDealProposal: *clientDealProposal,
+		State:              storagemarket.StorageDealUnknown,
+		Miner:              params.Info.PeerID,
+		MinerWorker:        params.Info.Worker,
+		DataRef:            params.Data,
+		FastRetrieval:      params.FastRetrieval,
+		DealStages:         storagemarket.NewDealStages(),
+		CreationTime:       curTime(),
+		IndexedCAR:         params.IndexedCAR,
```
raulk, line 386: I changed the name of this field to describe function, and not type. I don't think this affects existing deals, because I believe that CBOR is positional, but would love a confirmation, since you've seen several upgrades through @dirkmc

storagemarket/impl/provider.go
```diff
@@ -263,7 +262,7 @@ func (p *Provider) receiveDeal(s network.StorageDealStream) error {
 		Ref:                proposal.Piece,
 		FastRetrieval:      proposal.FastRetrieval,
 		CreationTime:       curTime(),
-		CARv2FilePath:      carV2FilePath,
+		InboundCAR:         path,
```
raulk, line 265: Changed the field name. Idem as above @dirkmc

storagemarket/impl/provider.go
```diff
@@ -595,22 +594,30 @@ func (p *Provider) start(ctx context.Context) error {
...
 	if err != nil {
 		return xerrors.Errorf("failed to fetch deals during startup: %w", err)
 	}
+
+	// re-track all deals for whom we still have a local blockstore.
+	for _, d := range deals {
+		if _, err := os.Stat(d.InboundCAR); err == nil && d.Ref != nil {
+			_, _ = p.stores.GetOrOpen(d.ProposalCid.String(), d.InboundCAR, d.Ref.Root)
+		}
+	}
```
raulk, line 614: We now load all in progress deals whose inbound CAR still exists on disk into the trackers. This reduces complexity of the former FinalizeReadWriteBlockstore on providerDealEnvironment, which used to accept the root CID and the path to reload the entry in the tracker if it didn't exist. It puzzled me quite a bit. I believe restoring state on startup is more correct.

storagemarket/impl/provider_environments.go
```diff
@@ -34,26 +34,37 @@ type providerDealEnvironment struct {
...
-func (p *providerDealEnvironment) CARv2Reader(carV2FilePath string) (*carv2.Reader, error) {
-	return carv2.OpenReader(carV2FilePath)
+func (p *providerDealEnvironment) ReadCAR(path string) (*carv2.Reader, error) {
+	return carv2.OpenReader(path)
 }
 
-func (p *providerDealEnvironment) FinalizeReadWriteBlockstore(proposalCid cid.Cid, carPath string, rootCid cid.Cid) error {
-	bs, err := p.p.readWriteBlockStores.GetOrCreate(proposalCid.String(), carPath, rootCid)
+func (p *providerDealEnvironment) FinalizeBlockstore(proposalCid cid.Cid) error {
+	bs, _, err := p.p.stores.Get(proposalCid.String())
```
raulk, line 45: Here's what I was talking about above.

stores/filestore.go
```diff
@@ -0,0 +1,161 @@
+package stores
```
raulk, line 1: The previous `filestorecaradapter` package is now merged into `stores`.

storagemarket/impl/client.go
dirkmc: I'm not 100% sure about that - maybe write a test to confirm?

dirkmc:
Looks like you need to run `make prepare-pr`

storagemarket/impl/provider_environments.go
```diff
@@ -34,26 +34,37 @@ type providerDealEnvironment struct {
...
 
+func (p *providerDealEnvironment) TerminateBlockstore(proposalCid cid.Cid, path string) error {
+	// stop tracking it.
+	if err := p.p.stores.Untrack(proposalCid.String()); err != nil {
+		log.Warnf("failed to clean read write blockstore, proposalCid=%s, car_path=%s: %s", proposalCid, path, err)
+	}
+
+	// delete the backing CARv2 file as it was a temporary file we created for
+	// this storage deal; the piece has now been handed off, or the deal has failed.
+	return os.Remove(path)
```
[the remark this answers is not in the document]
raulk: Agree. Will adjust!

stores/rw_bstores.go
```diff
@@ -0,0 +1,70 @@
...
+
+	"github.com/ipfs/go-cid"
+	"github.com/ipld/go-car/v2/blockstore"
+	"golang.org/x/xerrors"
+)
+
+type rwEntry struct {
+	bs   *blockstore.ReadWrite
+	path string
+}
```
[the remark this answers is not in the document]
raulk: We're actually not using it -- I thought we could but things ended up looking different. Good flag, I'll remove!

storagemarket/impl/client.go
raulk: It's gonna be hard writing a test for that :-( I'll hack something but it won't go in as a test.

storagemarket/impl/provider_environments.go
```diff
@@ -70,33 +81,23 @@ func (p *providerDealEnvironment) Ask() storagemarket.StorageAsk {
...
+	rd, err := carv2.OpenReader(carPath)
 	if err != nil {
-		return cid.Undef, "", xerrors.Errorf("failed to get CARv2 reader, proposalCid=%s, carV2FilePath=%s: %w", proposalCid, carV2FilePath, err)
+		return cid.Undef, "", xerrors.Errorf("failed to get CARv2 reader, proposalCid=%s, carPath=%s: %w", proposalCid, carPath, err)
 	}
 
 	defer func() {
 		if err := rd.Close(); err != nil {
-			log.Errorf("failed to close CARv2 reader, carV2FilePath=%s, err=%s", carV2FilePath, err)
+			log.Errorf("failed to close CARv2 reader, carPath=%s, err=%s", carPath, err)
```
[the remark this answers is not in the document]
raulk: I think this is OK since it's a log line.

raulk:
The `docs-check` CI target is broken in the CircleCI config, nothing to do with this PR.

PR #602, opened 2021-08-11
Replace client-side CAR file path with BlockstoreAccessor

dirkmc:
(Description added by @raulk)

This PR removes client-side CAR handling from go-fil-markets, and instead delegates on the calling code (Lotus in our immediate case) to provide blockstores on demand via an abstraction called `BlockstoreAccessor`.

Due to the lack of generics in Go, we have a `BlockstoreAccessor` for the storage market, and another for the retrieval market. The reason being that the keying varies (storage client: PayloadCID of the import, retrieval client: retrieval ID).


retrievalmarket/impl/clientstates/client_states.go
```diff
@@ -25,7 +26,7 @@ type ClientDealEnvironment interface {
 	OpenDataTransfer(ctx context.Context, to peer.ID, proposal *rm.DealProposal, legacy bool) (datatransfer.ChannelID, error)
 	SendDataTransferVoucher(context.Context, datatransfer.ChannelID, *rm.DealPayment, bool) error
 	CloseDataTransfer(context.Context, datatransfer.ChannelID) error
-	FinalizeBlockstore(context.Context, rm.DealID) error
+	FinalizeBlockstore(context.Context, cid.Cid) error
```
raulk, line 29: This is used with a PayloadCID. I don't think we should use it with a PayloadCID because there could be more than one retrieval in progress (from different miners) for the same payload CID.

hannahhoward (approved):
I'm good with this change. 

My only concern is if I wonder if we get into any circular order of operations rats nest where the dealID is created and returned, then a seperate thread tries to call stores.Get before the returned deal ID is passed and saved by the caller so it can return the store properly. Or maybe the accessor just makes stores on demand? I trust if you all have integrated in lotus and as long as it works fine there, I trust you to move forward.

PR #603, opened 2021-08-11
Replace car file path with BlockstoreAccessor in retrieval market client

dirkmc:


retrievalmarket/impl/client.go
```diff
@@ -28,11 +28,15 @@ import (
 	"github.com/filecoin-project/go-fil-markets/retrievalmarket/migrations"
 	rmnet "github.com/filecoin-project/go-fil-markets/retrievalmarket/network"
 	"github.com/filecoin-project/go-fil-markets/shared"
-	"github.com/filecoin-project/go-fil-markets/stores"
 )
 
 var log = logging.Logger("retrieval")
 
+type BlockstoreAccessor interface {
+	Get(key string, rootCid cid.Cid) (bstore.Blockstore, error)
```
raulk, line 36: I would make the argument a typed thing, and possibly introduce a type alias to make things super clear. Otherwise there's a chance that the implementation is storing CARs under different keys that they're being requested.

```
type PieceCID = cid.Cid
```

Also don't understand why the key _and the root CID?

retrievalmarket/impl/client.go
```diff
@@ -466,23 +463,8 @@ func (c *clientDealEnvironment) CloseDataTransfer(ctx context.Context, channelID
...
-	}
-
-	err = c.c.stores.Untrack(dealID.String())
-	if err != nil {
-		return xerrors.Errorf("failed to clean blockstore for deal with ID %s: %w", dealID, err)
-	}
-
-	return nil
+func (c *clientDealEnvironment) FinalizeBlockstore(ctx context.Context, payloadCid cid.Cid) error {
+	return c.c.bstores.Close(payloadCid.String())
```
raulk, line 467: Are we sure we want to ouse the payload CID here?

PR #608, opened 2021-08-19
Safe reader closer for PieceReaders

aarshkshah1992:
For https://github.com/filecoin-project/lotus/issues/7111.

As per @magik6k there is a small edge case in the JSON-RPC magic that can cause a read from the piece reader even after it's been closed by us after the AddPiece call returns. We should handle that gracefully.

### DIFF
--- a/storagemarket/impl/providerstates/safereadercloser.go
+++ b/storagemarket/impl/providerstates/safereadercloser.go
@@ -1,0 +1,40 @@
+package providerstates
+
+import (
+	"io"
+	"sync"
+
+	"golang.org/x/xerrors"
+)
+
+type safeReaderCloser struct {
+	r io.Reader
+	c io.Closer
+
+	lk sync.Mutex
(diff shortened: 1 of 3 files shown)


storagemarket/impl/providerstates/safereadercloser.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"golang.org/x/xerrors"
+)
+
+type safeReaderCloser struct {
+	r io.Reader
+	c io.Closer
+
+	lk       sync.Mutex
+	isClosed bool
```
raulk, line 15: You can simplify this to an atomic int32 and use the CAS (compare and swap primitives).

raulk:
Real solution is actually here: https://github.com/ipld/go-car/issues/211

aarshkshah1992:
Closing as per https://github.com/ipld/go-car/pull/212.

PR #619, opened 2021-08-28
rm go-multistore dependency.

raulk:
Farewell! Following the adoption of the dagstore, this library is no longer used.

`StoreID` fields have been kept for backwards compatibility and reset to uint64 (which should serialize identically to multistore.StoreID) but they're really unused.


hannahhoward (approved):
LGTM! Glad to move this forward

dirkmc (approved):

dirkmc:
Thanks! 👍

PR #629, opened 2021-09-08
retrieve by any CID (not just root CID) and reference provider integration

dirkmc:
- Retrieve by an cid not just the root cid.
- Reference provider integration in markets.

### TODO before Merge
- [x] Release dagstore and update deps.
- [x] Wait for a release of `indexer-provider` and update deps.
- [x] Wait for a release of https://github.com/filecoin-project/storetheindex and update deps.


retrievalmarket/impl/provider.go
```diff
@@ -109,6 +114,7 @@ func NewProvider(minerAddress address.Address,
 	dataTransfer datatransfer.Manager,
 	ds datastore.Batching,
 	retrievalPricingFunc RetrievalPricingFunc,
+	piecesWithBlock GetPiecesContainingBlock,
```
aarshkshah1992, line 117: Why don't we make this a method on the DAGStoreWrapper rather than adding on more abstraction here ?

retrievalmarket/impl/provider.go
```diff
@@ -388,48 +395,61 @@ func (p *Provider) HandleQueryStream(stream rmnet.RetrievalQueryStream) {
...
 }
 
-func (p *Provider) getPieceInfoFromCid(ctx context.Context, payloadCID, pieceCID cid.Cid) (piecestore.PieceInfo, bool, error) {
-	cidInfo, err := p.pieceStore.GetCIDInfo(payloadCID)
+// Given the CID of a block, find a piece that contains that block.
+// If the client has specified which piece they want, return that piece.
+// Otherwise prefer pieces that are already unsealed.
+func (p *Provider) getPieceInfoFromCid(ctx context.Context, payloadCID, clientPieceCID cid.Cid) (piecestore.PieceInfo, bool, error) {
+	// Get all pieces that contain the target block
+	piecesWithPayloadCID, err := p.piecesWithBlock(payloadCID)
```
aarshkshah1992, line 403: nit: variable name should be just `pieceCid`.

retrievalmarket/impl/provider.go
```diff
@@ -388,48 +395,61 @@ func (p *Provider) HandleQueryStream(stream rmnet.RetrievalQueryStream) {
...
+		return piecestore.PieceInfoUndefined, false, xerrors.Errorf("getting pieces for cid %s: %w", payloadCID, err)
 	}
+
+	// For each piece that contains the target block
 	var lastErr error
 	var sealedPieceInfo *piecestore.PieceInfo
-
-	for _, pieceBlockLocation := range cidInfo.PieceBlockLocations {
-		pieceInfo, err := p.pieceStore.GetPieceInfo(pieceBlockLocation.PieceCID)
+	for _, pieceWithPayloadCID := range piecesWithPayloadCID {
```
aarshkshah1992, line 411: `pieceCid`.

retrievalmarket/impl/provider.go
dirkmc: Before the function had a parameter called pieceCID, so I didn't want to also use pieceCID as the variable name for the cids of pieces containing the payload CID.
I think it's clearer if the two variables are named clientPieceCID and piecesWithPayloadCID

retrievalmarket/impl/provider.go
dirkmc: Makes sense 👍

retrievalmarket/impl/provider.go
dirkmc: Changed the name of piecesWithPayloadCID to piecesWithTargetBlock. Kind of a mouthful but I hope that's clearer :)

aarshkshah1992 (approved):

PR #632, opened 2021-09-28
add protocol to request retrieval ask

dirkmc:
Required by https://github.com/filecoin-project/lotus/issues/5820


nonsense (approved):

retrievalmarket/impl/provider.go
```diff
@@ -275,6 +276,42 @@ func (p *Provider) ListDeals() map[retrievalmarket.ProviderDealIdentifier]retrie
 	return dealMap
 }
 
+// HandleAskStream is called when the provider receives a request for the
+// retrieval ask (how much the provider charges for retrieval)
+func (p *Provider) HandleAskStream(s rmnet.RetrievalAskStream) {
```
aarshkshah1992, line 281: @dirkmc  This wont work. With Dynamic pricing, we quote a different price for each deal based on a number of factors. Just returning the ask from the ask store wont work. Maybe the storage provider has an external script configured for example. You want the client to give you the params it gives you when it actually makes the deal (pieceCid etc.) in the request for this stream and then run it through the same pricing function that we run during retrieval.

retrievalmarket/types.go
```diff
@@ -120,6 +124,12 @@ type RetrievalPeer struct {
 	PieceCID *cid.Cid
 }
 
+// SignedRetrievalAsk is an ask signed by the miner's private key
+type SignedRetrievalAsk struct {
```
aarshkshah1992, line 128: @dirkmc  We don't really sign the QueryResponse for the query protocol. So, why are we signing the ask response here ?

retrievalmarket/impl/askstore/askstore_impl.go
```diff
@@ -79,7 +99,40 @@ func (s *AskStoreImpl) GetAsk() *retrievalmarket.Ask {
 		return nil
 	}
 	ask := *s.ask
-	return &ask
+	return ask.Ask
+}
+
+// GetSignedAsk returns the current retrieval ask, signed with the provider's
+// key, or nil if there is no ask set.
+func (s *AskStoreImpl) GetSignedAsk() (*retrievalmarket.SignedRetrievalAsk, error) {
```
aarshkshah1992, line 107: Why do we need to sign asks ? Also, as commented elsewhere, just returning the ask here without using the pricing function wont be of much use.

aarshkshah1992 (changes requested):

retrievalmarket/impl/client.go
```diff
@@ -178,6 +178,56 @@ func (c *Client) FindProviders(payloadCID cid.Cid) []retrievalmarket.RetrievalPe
...
+	}
+
+	// Open a stream to the provider over the ask protocol
+	s, err := c.network.NewAskStream(ctx, p.ID)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to open stream to miner: %w", err)
+	}
+
+	// Send a request for the provider's ask
+	request := rmnet.AskRequest{Miner: p.Address}
```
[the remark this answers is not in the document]
dirkmc: We need to put _some_ parameter in the ask request in order to CBOR marshall it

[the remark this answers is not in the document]
dirkmc: I'll change this to the parameters needed for dynamic retrieval pricing

dirkmc:
Closing as this functionality doesn't add anything new over the `Query` protocol (see `HandleQueryStream`)

PR #633, opened 2021-10-01
Expose MaxTraversalLinks option for generating CommP

hannahhoward:
# Goals

Limit the depth of a DAG we're willing to process for CommP, expose this as
a configurable option via the public API.

Blockers:
- [x] tag go-ipld-prime v0.12.3 and update


rvagg:
I don't have review privs here so this is my +1 (for my own commit)

PR #639, opened 2021-10-07
chore(deps): use go-car master version

rvagg:
Ref: https://github.com/ipld/go-car/issues/254

Depends on getting null-padding added as an option to go-car master and getting a proper tag for this.


rvagg:
Now using the tagged, mainline v0.3.2 go-car, working just fine.
This is just blocked by https://github.com/filecoin-project/go-commp-utils/pull/7 being merged and tidied up here.
I'm confident at this stage that the functionality provided by the custom branch of go-car is no longer required because the code paths involved all go through go-car/v2 now which has the functionality available in its mainline releases.

rvagg:
We have a tagged go-car now and a new go-commp-utils, so this is now just a version bump and clean up of modules, I think this can be merged now (although I had to rerun CI a few times to pass the flakys).

mvdan:
> Depends on getting null-padding added as an option to go-car master and getting a proper tag for this.

Just to clarify, this line from the original post is no longer relevant, right? As you no longer think "zero as EOF" is needed for the uses of the go-car v0 module.

mvdan (approved):
Assuming we're all in agreement the two unmerged go-car v0 commits aren't actually needed, the changes LGTM!

rvagg:
replaced with #709, this PR was made redundant by #653 anyway

PR #640, opened 2021-10-07
Standardize definition of a recursive non-matching selector

ribasushi:
In the spirit of https://github.com/filecoin-project/lotus/pull/7467

It was confirmed that [key order in this map](https://github.com/filecoin-project/lotus/pull/7467#issuecomment-937664000) is irrelevant, therefore this is a non-functional change.

### DIFF
--- a/shared/selectors.go
+++ b/shared/selectors.go
@@ -2,15 +2,9 @@ package shared
 
 import (
 	"github.com/ipld/go-ipld-prime"
-	basicnode "github.com/ipld/go-ipld-prime/node/basic"
-	"github.com/ipld/go-ipld-prime/traversal/selector"
-	"github.com/ipld/go-ipld-prime/traversal/selector/builder"
+	selectorparse "github.com/ipld/go-ipld-prime/traversal/selector/parse"
 )
 
-// entire DAG selector
-func AllSelector() ipld.Node {
-	ssb := builder.NewSelectorSpecBuilder(basicnode.Prototype.Any)
-	return ssb.ExploreRecursive(selector.RecursionLimitNone(),
-		ssb.ExploreAll(ssb.ExploreRecursiveEdge())).
-		Node()
-}
+// AllSelector is a compatibility alias for an entire DAG non-matching-selector
+// Use selectorparse.CommonSelector_ExploreAllRecursively directly in new code
+func AllSelector() ipld.Node { return selectorparse.CommonSelector_ExploreAllRecursively }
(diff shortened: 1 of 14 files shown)


shared/selectors.go
```diff
@@ -2,15 +2,9 @@ package shared
...
-// entire DAG selector
-func AllSelector() ipld.Node {
-	ssb := builder.NewSelectorSpecBuilder(basicnode.Prototype.Any)
-	return ssb.ExploreRecursive(selector.RecursionLimitNone(),
-		ssb.ExploreAll(ssb.ExploreRecursiveEdge())).
-		Node()
-}
+// AllSelector is a compatibility alias for an entire DAG non-matching-selector
+// Use selectorparse.CommonSelector_ExploreAllRecursively directly in new code
+func AllSelector() ipld.Node { return selectorparse.CommonSelector_ExploreAllRecursively }
```
rvagg, line 10: can't we just delete this now or are we trying to do a gentle deprecation?

mvdan: Yeah I'd say keep it and mark it deprecated: https://github.com/golang/go/wiki/Deprecated

There's nearly zero cost to doing that, given how tiny the code is.

ribasushi: @mvdan could you do a "code-suggestion" right here? I am not well versed in the right etiquette for deprec. comments...

mvdan: Also, FYI, you can also deprecate the entire package.

ribasushi: > you can also deprecate the entire package.

mmmm it has other functions ( different files ), this doesn't apply here does it?

shared/selectors.go
```diff
@@ -2,15 +2,9 @@ package shared
...
 
-// entire DAG selector
-func AllSelector() ipld.Node {
-	ssb := builder.NewSelectorSpecBuilder(basicnode.Prototype.Any)
-	return ssb.ExploreRecursive(selector.RecursionLimitNone(),
-		ssb.ExploreAll(ssb.ExploreRecursiveEdge())).
-		Node()
-}
+// AllSelector is a compatibility alias for an entire DAG non-matching-selector
+// Use selectorparse.CommonSelector_ExploreAllRecursively directly in new code
```
mvdan, line 9:
```suggestion
// Deprecated: AllSelector is a compatibility alias for an entire DAG non-matching-selector.
// Use github.com/ipld/go-ipld-prime/traversal/selector/parse.CommonSelector_ExploreAllRecursively instead.
```

I think spelling out the full import path is valuable, because it's a different module entirely, so the user might have a really hard time figuring out what package `selectorparse` or `selector/parse` is.

dirkmc (approved):
LGTM 👍 

Thanks!

PR #645, opened 2021-10-20
use retrieval ask interface as impl argument

hunjixin:


hunjixin:
this pr aim to allow user use other db for ask read/write.

PR #650, opened 2021-12-06
Update go-car to v2.1.0 

magik6k:
This PR updates go-car to v2.1.0, and fixes compat with the filestore car(ish) index in the least hacky way per Slack discussion.


mvdan:
In kvcarbs.go, perhaps clarify what lines you changed from upstream and how, and state what commit hash it came from. For example, perhaps comment out a few lines as necessary, or do an `if false { ... }`. That should make the changes clearer, and if the code needs to be updated to a newer go-car in the future, at least it's going to be possible without pulling you in.

dirkmc (approved):

PR #652, opened 2021-12-07
Fix and de-duplicate link to spec

NiloCK:
The prior link forwarded correctly from the gh pages link to spec.filecoin.io, but did not land the user at the correct section.


dirkmc (approved):
Thanks! 👍

PR #655, opened 2021-12-17
Timeout provider after waiting a period of time for transfer to restart

hannahhoward:
# Goals

Currently, when a provider restarts when a data transfer is in progress, it goes into a waiting state for the client to restart the transfer. If the client doesn't restart the transfer, the provider simply stays in this state forever, never failing the deal. This can happen for example if the provider was offline long enough for the client to attempt restarts until it gave up and failed the deal on its own side.

# Implementation

After a configurable amount of time (default 1 hour) in the awaiting restart state, if the transfer is not restart, fail the deal. This is accomplished via a state handler that fires a timeout event always after a the given time. If the state has transitioned by then, the event has no effect. If it has not transitioned to something else, and the provider is still waiting, the deal fails.


storagemarket/impl/providerstates/provider_states.go
```diff
@@ -208,6 +210,23 @@ func DecideOnProposal(ctx fsm.Context, environment ProviderDealEnvironment, deal
 	return ctx.Trigger(storagemarket.ProviderEventDataRequested)
 }
 
+const restartTimeout = 1 * time.Hour
```
dirkmc, line 213: Is `restartTimeout` used anywhere?

dirkmc (approved):
One nit but otherwise LGTM 👍

storagemarket/impl/providerstates/provider_states.go
hannahhoward: nope accidental leftover -- became defaultRestartTimeout -- will delete

PR #657, opened 2021-12-20
fix storage provider restart in publish stage

dirkmc:
Solution to https://github.com/filecoin-project/lotus/issues/7815

### DIFF
--- a/storagemarket/impl/provider.go
+++ b/storagemarket/impl/provider.go
@@ -614,13 +614,6 @@ func (p *Provider) runMigrations(ctx context.Context) ([]storagemarket.MinerDeal
 		return nil, xerrors.Errorf("failed to fetch deals during startup: %w", err)
 	}
 
-	// re-track all deals for whom we still have a local blockstore.
-	for _, d := range deals {
-		if _, err := os.Stat(d.InboundCAR); err == nil && d.Ref != nil {
-			_, _ = p.stores.GetOrOpen(d.ProposalCid.String(), d.InboundCAR, d.Ref.Root)
-		}
-	}
-
 	// migrate deals to the dagstore if still not migrated.
 	if ok, err := p.dagStore.MigrateDeals(ctx, deals); err != nil {
 		return nil, fmt.Errorf("failed to migrate deals to DAG store: %w", err)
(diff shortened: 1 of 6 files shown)


storagemarket/impl/provider.go
```diff
@@ -614,13 +614,6 @@ func (p *Provider) runMigrations(ctx context.Context) ([]storagemarket.MinerDeal
 		return nil, xerrors.Errorf("failed to fetch deals during startup: %w", err)
 	}
 
-	// re-track all deals for whom we still have a local blockstore.
-	for _, d := range deals {
-		if _, err := os.Stat(d.InboundCAR); err == nil && d.Ref != nil {
-			_, _ = p.stores.GetOrOpen(d.ProposalCid.String(), d.InboundCAR, d.Ref.Root)
-		}
-	}
-
```
dirkmc, line 623: This piece of code was causing the problem - we shouldn't need to do this as stores.GetOrOpen() will be called lazily

storagemarket/impl/provider_environments.go
```diff
@@ -107,36 +108,8 @@ func (p *providerDealEnvironment) GeneratePieceCommitment(proposalCid cid.Cid, c
 		}
 	}()
 
-	// dump the CARv1 payload of the CARv2 file to the Commp Writer and get back the CommP.
```
dirkmc, line 110: Moved this code to its own file so it can be used from the test - no logical changes

storagemarket/testnodes/testnodes.go
```diff
@@ -380,7 +382,28 @@ func (n *FakeProviderNode) WaitForPublishDeals(ctx context.Context, mcid cid.Cid
...
+	}
+
+	// We read in all the bytes from the reader above, so seek back to the start
+	err := pieceReader.SeekStart()
+	if err != nil {
+		return nil, fmt.Errorf("on deal complete: seeking to start of piece data: %w", err)
+	}
+
+	// Generate commP
+	pieceCID, err := shared.GenerateCommp(pieceReader, uint64(pieceSize), uint64(pieceSize))
```
dirkmc, line 397: I added commp generation to the test so if the data is bad the test will fail

magik6k (approved):
Looks reasonable to me

PR #663, opened 2022-01-15
Update for go-data-transfer v1.13.0

hannahhoward:
# Goals

Support for go-data-transfer v1.13.0 which accepts transfers of incomplete DAGs that are missing CIDs

# Implementation

The intent is two fold:
1. A Storage deal transfer that is missing CIDs should just succeed. If someone assembles a CAR that is missing CIDs, as long as it matches commP on the other side we should accept the deal. Note: the CAR tooling needs updates as well.
2. A Retrieval deal on the client records when it is missing cids -- this can be presented to the user at the end of retrieval as needed.

# For Discussion

The intent is for this to get into Lotus 1.15. However, I see all the reference provider work has been done, causing me to wonder if this needs a markets 1.14 backport.


dirkmc (approved):

dirkmc:
Superseded by https://github.com/filecoin-project/go-fil-markets/pull/665

PR #665, opened 2022-01-18
 Update for go-data-transfer v1.13.0

dirkmc:
This PR branches from before the v1.15.0 tag was created, so that we can create a version of go-fil-markets with the new go-data-transfer and graphsync versions, but without the changes in https://github.com/filecoin-project/go-fil-markets/pull/629

The original version of this PR (that branches from after the v1.15.0 tag) is at https://github.com/filecoin-project/go-fil-markets/pull/663


dirkmc:
Closing as we are no longer going to implement these changes

PR #673, opened 2022-01-20
Index publishing work

aarshkshah1992:


retrievalmarket/impl/client_test.go
```diff
@@ -225,6 +225,7 @@ func TestClient_FindProviders(t *testing.T) {
 // retrieval deal for the same payload CID with the same peer as an existing
 // active deal
 func TestClient_DuplicateRetrieve(t *testing.T) {
+	t.Skip("flaky test")
```
hannahhoward, line 228: Did it become more flaky in this PR?
1. Yes -- then let's figure that out
2. No -- I would not throw in a skip as part of a feature PR -- i'd just make an issue and if we can't tackle it now, let's make a seperate PR to add a skip

retrievalmarket/impl/provider.go
```diff
@@ -448,6 +461,65 @@ func (p *Provider) pieceInUnsealedSector(ctx context.Context, pieceInfo piecesto
 	return false
 }
 
+func (p *Provider) storageDealsForPiece(clientSpecificPiece bool, payloadCID cid.Cid, pieceInfo piecestore.PieceInfo) ([]abi.DealID, error) {
```
hannahhoward, line 464: This is an existing problem (not introduced by this PR) but this method has a weird logic to it -
- if the first parameter is true, then the third parameter is used, the second parameter unused
- if the first parameter is false, the second parameter is used, the third parameter unused

To me this says: make it two methods -- and also, consider the interrelated logic from the the HandleQueryStream/getPieceInfoFromCid and see if there's a way to simply.

Definitely non-blocking, just noting.

storagemarket/impl/provider.go
```diff
@@ -177,6 +190,25 @@ func (p *Provider) Start(ctx context.Context) error {
 			log.Error(err.Error())
 		}
 	}()
+
+	// connect the index provider node with the full node and protect that connection
```
hannahhoward, line 194: This feels like a very odd concern to put in go-fil-markets... maybe I don't know the context? I would expect it to live in the Lotus markets process. (or later Boost)

storagemarket/impl/providerstates/provider_states.go
```diff
@@ -390,6 +392,15 @@ func HandoffDeal(ctx fsm.Context, environment ProviderDealEnvironment, deal stor
 		log.Error(err)
 	}
 
+	// announce the deal to the network indexer
+	annCid, err := environment.AnnounceIndex(ctx.Context(), deal)
+	if err != nil {
+		log.Errorw("failed to announce index via reference provider", "proposalCid", deal.ProposalCid, "err", err)
```
hannahhoward, line 398: Should the fact that an error occurred in publish be saved to the deal state? Other than logs, how do I know which of my deals got announced successfully? Do I always want AnnounceAllDealsToIndex() to announce everything?

hannahhoward (commented):
My biggest meta comment is you should remove:
- CIDInfo entirely from the PieceStore -- it's unused
- Remove the whole "MetadataPath" if that machinery still exists -- it's also unused

It's important to make clear to someone reading the code that the single source of truth on which payloadCIDs are in which pieces is now in the DagStore only.

storagemarket/impl/providerstates/provider_states.go
aarshkshah1992: @hannahhoward The next time a deal gets announced, because of the IPLD magic in the index provider, it will also announce all the previous un-announced deal when the network indexer asks for the latest.

retrievalmarket/impl/client_test.go
aarshkshah1992: 2. Issue at https://github.com/filecoin-project/go-fil-markets/issues/683.

storagemarket/impl/providerstates/provider_states.go
masih: @aarshkshah1992 not necessarily. 
If the error happens because the advertisement failed to persist then that content will never get advertised. 
Looking at the code, such errors can hapen before call to `NotifyPut` e.g. in metadat encoding on markets side or inside the provider before persistence.

The case you mentioned above will only work if the new advertisement is added to the chain but not announced to the network due to say gossipsub error.

aarshkshah1992:
> My biggest meta comment is you should remove:
> 
> * CIDInfo entirely from the PieceStore -- it's unused
> * Remove the whole "MetadataPath" if that machinery still exists -- it's also unused

It's not really a part of this work. I started doing thus but looks like a bigger refactor given that it's very pervasive in all the tests too. I've created https://github.com/filecoin-project/go-fil-markets/issues/684 to track this and we can do this in a follow up PR.

> It's important to make clear to someone reading the code that the single source of truth on which payloadCIDs are in which pieces is now in the DagStore only.

Have added a comment for this.

storagemarket/impl/providerstates/provider_states.go
masih: https://github.com/filecoin-project/go-fil-markets/issues/685

PR #675, opened 2022-02-02
Connect index provider host to full node

aarshkshah1992:
Ensure the index provider host is connected to the full node so that it's a part of the gossipsub mesh.


storagemarket/impl/provider.go
```diff
@@ -184,6 +193,25 @@ func (p *Provider) Start(ctx context.Context) error {
...
+			select {
+			case <-time.After(time.Minute):
+				if err := p.connectIndexProviderToFullNode(ctx); err != nil {
+					log.Errorf("failed to connect index provider host with the full node: %s", err)
+				}
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
```
nonsense, line 213: AFAICT this is a responsibility of the libp2p connection manager (i.e. `protected peers`). Are we sure that we need this piece of code, and it is not redundant?

aarshkshah1992: Hey Anton, this is to ensure we keep connecting if something goes wrong and the full node disconnects us. It's just there for safety. If we already have a connection to the full node, this will be a no-op as it dosen't create another connection.

nonsense (approved):

PR #707, opened 2022-05-03
Voucher refactor integration

hannahhoward:
# Goals

This PR rewrites the go-fil-markets retrieval provider based on the new data transfer voucher refactor (https://github.com/filecoin-project/go-data-transfer/pull/308, https://github.com/filecoin-project/go-data-transfer/pull/322) as well as the new go-statemachine multi-event application PR (https://github.com/filecoin-project/go-statemachine/pull/37).

# Implementation

The retrieval client code is unchanged for now -- this change is intended to be backward compatible for the time being -- even though we could seriously drop a ton of code if we stopped having the client wait for the provider to request payment.

All of the revalidation code is -> gone (hence the big delete/add ratio)

The validation code is mostly the same, but we set data limits + pauses  + require finalization

Restarts are now handled seperately, and we handle them by simply looking at the current state of the transfer and setting  validation status appropriately.

When a data limit or finalization occurs, we trigger the payment states, which in turn trigger sending an updated validation status + a voucher result that requests payment.

When a new voucher arrives, we trigger the payment states again, which processes the last payment and updates the funds received as needed, then sets validation status based on that.

The long and short of all this is we no longer do any tracking of transfer state directly in the provider -- only funds received, the unique state to this layer.

I'm also dropping all the v0 protocols, which haven't actually been in use since pre-mainnet.


retrievalmarket/impl/provider.go
```diff
@@ -92,13 +91,6 @@ func DealDeciderOpt(dd DealDecider) RetrievalProviderOption {
 	}
 }
 
-// DisableNewDeals disables setup for v1 deal protocols
```
rvagg, line 95: I'm not familiar with the v0/v1 deal proposal thing, but what are the implications of removing the ability to disable this? Is v0 ancient by now and we don't foresee any need to provide the optionality on opting in to v0-only at this stage?

rvagg: ah, I just re-read the OP:

> I'm also dropping all the v0 protocols, which haven't actually been in use since pre-mainnet.

so nevermind then; I see that v0 code is also being dropped completely with no migration path. Fair enough if it's as ancient as pre-mainnet.

retrievalmarket/impl/providerstates/provider_states.go
```diff
@@ -39,40 +41,125 @@ func UnsealData(ctx fsm.Context, environment ProviderDealEnvironment, deal rm.Pr
...
+			log.Debugf("provider: funds needed: last payment")
+			voucherResult = &rm.DealResponse{
+				ID:     deal.ID,
+				Status: rm.DealStatusCompleted,
+			}
+		}
+	}
+	return datatransfer.ValidationResult{
+		Accepted:             true,
+		VoucherResult:        voucherResult,
```
rvagg, line 113: it's ok that this could be an empty voucherResult?

retrievalmarket/types.go
```diff
@@ -260,15 +249,71 @@ func (p Params) IntervalLowerBound(currentInterval uint64) uint64 {
...
 		lowerBound = target
 		target += intervalSize
 		intervalSize += p.PaymentIntervalIncrease
 	}
 	return lowerBound
 }
 
-func (p Params) NextInterval(currentInterval uint64) uint64 {
+// OutstandingBalance produces the amount owed based on the deal params
+// for the given transfer state and funds received
```
rvagg, line 261: "given transfer state" - so I see that this method returns the oustanding unsealed price as the "outstanding balance" but if the unsealing price is covered then the "outstanding balance" is something different. So this is really an "OutstandingBalanceForState", is that correct? (no need to rename, you say this in the comment, I'm just making sure I'm reading this right).

retrievalmarket/impl/providerstates/provider_states.go
hannahhoward: yes. Nil is an acceptable voucher result

retrievalmarket/impl/provider.go
hannahhoward: yea, I'm going to do an additional PR to drop the on disk storage portion as well... I have no idea why it lived this long.

retrievalmarket/types.go
hannahhoward: yes. amount transferred will always be zero (or should always be zero) if the unseal price hasn't been paid, as unsealing won't have even begun, which means that transfer will not have begun. :)

I'll do some commenting here.

rvagg:
@hannahhoward 

> this change is intended to be backward compatible

One minor concern is that `ProviderDealState` is changing shape: `TotalSent`, `CurrentInterval` and `LegacyProtocol` are all gone now and they're not optional fields. I think this struct is only used for saving state and that cbor-gen should gracefully ignore these (both in and out) and I'm not entirely clear on how the statemachine saving stuff works so it'd be good to hear your confirmation that you think this is OK.

PR #709, opened 2022-05-10
chore: update go-car dependency

rvagg:
Ref: #639
Ref: https://github.com/ipld/go-car/issues/254

Backstory: once upon a time, go-car had a branch that was used here to act as a filestore, but that branch was never merged to master. Over time, go-fil-markets evolved away from that particular usage of go-car, eventually making it disappear entirely in favour of go-car/v2.

https://github.com/ipld/go-car/issues/254 documents attempts to get rid of this branch for good.

#639 was a previous attempt to start removing this branch dependency here.

The last update to the go-car dependency here was #653 which just replaced the branch ref with the latest master, which worked fine because by that time the use of the filestore and need for zero==eof was redundant.

So now we're just moving to a tagged version here and ticking off an item back in https://github.com/ipld/go-car/issues/254
.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -44,7 +44,7 @@ require (
 	github.com/ipfs/go-log/v2 v2.5.0
 	github.com/ipfs/go-merkledag v0.5.1
 	github.com/ipfs/go-unixfs v0.3.1
-	github.com/ipld/go-car v0.3.3-0.20211210032800-e6f244225a16
+	github.com/ipld/go-car v0.3.3
 	github.com/ipld/go-car/v2 v2.1.1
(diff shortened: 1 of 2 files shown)


rvagg:
TestRestartClient/ClientEventDealPublished is failing in CI and has been failing locally for me, even prior to this change, I don't have background on that and whether it's an expected flaky or not but I don't believe it's related to this.

hannahhoward (approved):

hannahhoward:
honestly don't worry about the test failure. though wow we should really get our circle updated so we can at least run tests. Eek!

hannahhoward:
I'm just gonna invoke those sweet admin privileges and merge

PR #710, opened 2022-05-11
fix:close ask/dealstatus stream

hunjixin:
close the ask/dealstatus stream, which will cause libp2p resouce resource over limit


dirkmc (approved):

dirkmc:
Thanks! 🙏

PR #711, opened 2022-05-12
Use new go-state-types state accessors

arajasek:
Replaces #705 .

I'm not sure what's causing the test failure, and can't repro it locally. @ZenGround0 would you be able to take a look, I'm a little stuck.


retrievalmarket/migrations/migrations_test.go
```diff
@@ -10,19 +10,25 @@ import (
...
 	"github.com/filecoin-project/go-statemachine/fsm"
-	tutils "github.com/filecoin-project/specs-actors/v8/support/testing"
 
 	"github.com/filecoin-project/go-fil-markets/piecestore/migrations"
 	"github.com/filecoin-project/go-fil-markets/retrievalmarket"
 	"github.com/filecoin-project/go-fil-markets/retrievalmarket/impl/clientstates"
 	"github.com/filecoin-project/go-fil-markets/retrievalmarket/impl/providerstates"
 )
 
+func NewActorAddr(t testing.TB, data string) address.Address {
```
ZenGround0, line 26: Ditto

ZenGround0 (approved):

PR #721, opened 2022-06-17
migrate provider deal proposal label to the new format

dirkmc:
I made a build of v1.16.0-rc1 with go-fil-markets updated to point at this PR.
I tested that I could make a storage deal and it published successfully on my local devnet with the following combinations of client and provider:
- [x] client v1.15.3 / provider v1.15.3 ✅ 
- [x] client v1.15.3 / provider v1.16.0-rc1 ✅
- [x] client v1.16.0-rc1 / provider v1.15.3 ✅
- [x] client v1.16.0-rc1 / provider v1.16.x ✅


storagemarket/impl/provider.go
```diff
@@ -265,14 +265,9 @@ func (p *Provider) receiveDeal(s network.StorageDealStream) error {
 		return xerrors.Errorf("failed to get deal proposal from proposal message")
 	}
 
-	proposalNd, err := cborutil.AsIpld(proposal.DealProposal)
-	if err != nil {
-		return err
-	}
-
```
Stebalien, line 272: I'm not seeing why this change was necessary. Unless it's just cleanup.

storagemarket/network/deal_stream_v110.go
```diff
@@ -0,0 +1,82 @@
...
+		log.Warnf(err.Error())
+		return ProposalUndefined, cid.Undef, err
+	}
+
+	prop, err := migrations.MigrateClientDealProposal0To1(*ds.DealProposal)
+	if err != nil {
+		err = fmt.Errorf("migrating v110 deal proposal to current version: %w", err)
+		log.Warnf(err.Error())
+		return ProposalUndefined, cid.Undef, err
+	}
```
Stebalien, line 47: Is this from the user? We can't do this migration without (possibly) changing signatures.

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/filecoin-project/go-fil-markets
 go 1.13
 
 require (
+	github.com/btcsuite/btcd/chaincfg/chainhash v1.0.1 // indirect
```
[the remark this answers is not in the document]
magik6k: This breaks boost build for me; Removing this like fixes it
```
github.com/filecoin-project/boost/cli/node imports
        github.com/libp2p/go-libp2p-core/crypto imports
        github.com/btcsuite/btcd/btcec tested by
        github.com/btcsuite/btcd/btcec.test imports
        github.com/btcsuite/btcd/chaincfg/chainhash: ambiguous import: found package github.com/btcsuite/btcd/chaincfg/chainhash in multiple modules:
        github.com/btcsuite/btcd v0.22.0-beta (/home/magik6k/.opt/go/pkg/mod/github.com/btcsuite/btcd@v0.22.0-beta/chaincfg/chainhash)
        github.com/btcsuite/btcd/chaincfg/chainhash v1.0.1 (/home/magik6k/.opt/go/pkg/mod/github.com/btcsuite/btcd/chaincfg/chainhash@v1.0.1)
```

jennijuju:
(this needs to be backported on top of ` v1.20.1-v16-1` as well.

storagemarket/impl/provider.go
dirkmc: The provider deal datastore is indexed by signed deal proposal cid.
If the Label field contains bytes that are not a utf8-compliant string, the CBOR-marshaled deal proposal will be different between the old format (label as a string) and the new format (DealLabel)
Therefore the signed deal proposal cid will also be different.

Lotus client
- generates the signed deal proposal cid
- sends a deal proposal
- reads the response
- checks that the signed deal proposal cid returned by the provider matches the one generated on the client

When the provider receives a deal proposal we need to know what format the deal proposal was in when the client CBOR-marshaled it, in order to be able to create the right signed deal proposal cid. `ReadDealProposal` checks the deal proposal protocol version to determine how to interpret the CBOR-marshaled deal proposal (old format vs new format)

storagemarket/network/deal_stream_v110.go
dirkmc: Good catch 👍 

One option would be to reject non-utf8 labels sent over version 1.1.0 of the deal proposal protocol.

go.mod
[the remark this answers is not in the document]
dirkmc: This is how the libp2p team fixed it:
https://github.com/libp2p/go-libp2p-core/pull/254

storagemarket/impl/provider.go
dirkmc: I backed out this change - instead we reject deal proposals over the v1.1.0 protocol where the deal label is not parseable utf8

storagemarket/network/deal_stream_v110.go
dirkmc: After discussing with @magik6k and @arajasek I implemented the proposed change: reject deal proposals over the v1.1.0 protocol where the deal label is not parseable utf8

arajasek: @dirkmc just for my sanity, this will only affect any new deals made, not historical deals that may have non-utf8 labels, yes?

arajasek (approved):

storagemarket/network/deal_stream_v110.go
dirkmc: Historical deals will be migrated from the old format (deal label as a string) to the new format (DealLabel struct).
If those deals were made with a non-utf8 deal label, their signature will no longer match the new format. It's my understanding that this doesn't matter because those deals have already been published on chain.

This PR introduces a new protocol version. So we now have:
- v1.1.0: The current version
  When a deal proposal is received, reject the deal proposal if it has a label that cannot be parsed as utf8.
  Otherwise, migrate the deal proposal to the new format (label as DealLabel struct). This should not change the signature of the deal because the label is CBOR-marshalled in the same was as if it were a string. See the tests in [storagemarket/network/deal_stream_v110_test.go](https://github.com/filecoin-project/go-fil-markets/pull/721/files#diff-300884c916912bfde5ed40e33b85b32560a88e8cb3af07d4dea76be1ebdfe20d)
- v1.1.1: The new version
  This protocol version will be used when both the client and provider use a version of lotus that depends on this PR.
  Deals made over v1.1.1 are already in the new format (label as a DealLabel struct).

go.mod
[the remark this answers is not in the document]
dirkmc: Merged the above PR with a fix into this branch.

storagemarket/network/deal_stream_v110.go
arajasek: @dirkmc Got it, I think this is good for any newly made / received deals. My concern is whether the entire migration of historical deals will fail if there are historical non-utf8 deals (which I agree will no longer have valid signatures) -- and if so, what is the impact of this migration failing?

(also when does the migration of historical deal info happen? at startup?)

dirkmc: The migration happens at startup 👍 

The migration is not expected to fail for historical deals with non-utf8 labels. Please take a look at the code here:
https://github.com/filecoin-project/go-fil-markets/pull/721/files#diff-8536cd1664598e5c06e27807ae8fa95a9f1d40439d4883fbef351e54826dbeccR248
If there's an issue with that code let me know.

arajasek: Perfect, thanks, that assuages my concern -- I think I was conflating the `ReadDealProposal` (which uses `checkDealLabel` and fails if non-utf-8) with the migration code, that succeeds in an immaterially wrong state.

Thanks for indulging!

dirkmc: Ah yep, that makes sense 👍

PR #738, opened 2022-07-26
Include pieceinfo for deal decider

hannahhoward:
# Goal

A minor changes means that when custom deal decider is called, it includes deal.PieceInfo, which has the found piece CID when none is passed in the deal proposal


bhndt (approved):

bhndt (approved):
add

rvagg (approved):

PR #739, opened 2022-07-31
 use ready manager for retrieval provider

hannahhoward:
# Goals

switch to the ready manager as opposed to raw pubsub for ready in retrieval provider


rvagg (approved):
👍 it doesn't look like there's a risk here of duplicate events being fired but I suppose the more focused tool is the right way to do this.

PR #749, opened 2022-09-02
fix: store identity CIDs in CARs for online deals

rvagg:
Builds on #747 and relies on https://github.com/ipld/go-car/pull/332

This matches historical precedent, without doing this we get CommP failures where
deals are made with intermediate inline CIDs and the client stores them in
the CAR used to make CommP. car.SelectiveCar is the historical way of doing
this.

e.g. https://github.com/filecoin-project/lotus/blob/a843c52e38da13da489cbe6b290ea49b2660b3fb/node/impl/client/client.go#L1405-L1412

More discussion in https://github.com/ipld/go-car/pull/332, and I'll also open a Lotus PR to bring this in and test it there and will link back here from that.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -39,7 +39,7 @@ require (
 	github.com/ipfs/go-merkledag v0.6.0
 	github.com/ipfs/go-unixfs v0.3.1
 	github.com/ipld/go-car v0.4.0
-	github.com/ipld/go-car/v2 v2.4.1
+	github.com/ipld/go-car/v2 v2.4.2-0.20220902111941-7b1af9af7fe8
 	github.com/ipld/go-codec-dagpb v1.3.1
 	github.com/ipld/go-ipld-prime v0.17.0
 	github.com/jbenet/go-random v0.0.0-20190219211222-123a90aedc0c
--- a/go.mod
+++ b/go.mod
@@ -39,7 +39,7 @@ require (
 	github.com/ipfs/go-merkledag v0.6.0
 	github.com/ipfs/go-unixfs v0.3.1
 	github.com/ipld/go-car v0.4.0
-	github.com/ipld/go-car/v2 v2.4.2-0.20220902111941-7b1af9af7fe8
+	github.com/ipld/go-car/v2 v2.5.0
 	github.com/ipld/go-codec-dagpb v1.3.1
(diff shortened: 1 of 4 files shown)


rvagg:
updated this to use the newly released go-car/v2@2.5.0 which has the necessary changes to handle identity CIDs like we want for this

retrievalmarket/impl/provider.go
```diff
@@ -390,62 +414,180 @@ func (p *Provider) HandleQueryStream(stream rmnet.RetrievalQueryStream) {
...
-			if sealedPieceInfo == nil {
-				// The piece is not in an unsealed sector, so save it but keep
-				// checking other pieces to see if there is one that is in an
-				// unsealed sector
-				sealedPieceInfo = &pieceInfo
+	pieces := make([]cid.Cid, 0)
+	// for each link, query the dagstore for pieces that contain it
+	for i, link := range links {
+		piecesWithThisCid, err := p.dagStore.GetPiecesContainingBlock(link)
+		if len(piecesWithThisCid) == 0 {
```
dirkmc, line 504: Is it possible for `GetPiecesContainingBlock` to return a zero-length piecesWithThisCid, and for err to also be nil?

dirkmc (commented):
Just one question, but otherwise LGTM 👍 

There is quite a lot of code for handling pieces in provider.go
I wonder if we should move some of it into a separate file?

rvagg:
This becomes _considerably_ simpler with #747 merged, the big changes were in there. Rebased now so there's only two commits left.

dirkmc (approved):

retrievalmarket/impl/provider.go
rvagg: Good call, I think it's probably unsafe to not explicitly deal with `err` even though I believe zero-length should also ways be coupled with an error (comes out of `GetShardsForMultihash` in the inverted index which requires an entry for that CID and an ErrNotFound otherwise). Addressed in https://github.com/filecoin-project/go-fil-markets/pull/750

rvagg:
> There is quite a lot of code for handling pieces in provider.go
> I wonder if we should move some of it into a separate file?

Done in https://github.com/filecoin-project/go-fil-markets/pull/750 since those changes related to the now-merged #747

PR #773, opened 2022-12-31
stop relying on the deprecated go-libp2p-core package

marten-seemann:
We moved all of the go-libp2p-core packages into go-libp2p. This makes future updates a lot easier, but it means we'll have to adjust import paths once now.


neelvirdy:
@marten-seemann are you still tracking this? running into blockers updating dependencies on filclient and this would help unblock me

marten-seemann:
This PR should be good to go? Looks like it needs to be rebased though. Want to take it over?

neelvirdy:
sure, recreated here https://github.com/filecoin-project/go-fil-markets/pull/777

PR #774, opened 2023-01-09
Upgrade to the latest index-provider `v.10.0`

masih:
Upgrade to the latest index-provider library; note that the library is moved to `ipni` GitHub org and its go module is renamed, replacing `filecoin-project` with `ipni`.


willscott (approved):

willscott:
Don't trust my approval though - check with @dirkmc  on which branch is the right one for this

PR #789, opened 2023-04-01
Chore/use go libipni

gammazero:


rvagg (commented):
So, these look good to me, I love dependency upgrades, especially this lot!

But, the failing tests look sus and I don't know much about them to be able to make a judgement call on them. They're all `TestBounceConnection*` and are still failing after a re-run. I think this is on @dirkmc to make a call on.

dirkmc:
Those tests are flaky - if they pass locally I think it's fine to merge this PR

gammazero:
@dirkmc With the latest graphsync updates, these tests pass now.

dirkmc (approved):

PR #796, opened 2023-08-16
Add json marshalling compatibility

hannahhoward:
# Goals

add compatibility interfaces to CborGenCompatibleNode for encoding/json so it can be easily mashalled/unmarshalled to JSON

### DIFF
--- a/retrievalmarket/bindnodeoptions.go
+++ b/retrievalmarket/bindnodeoptions.go
@@ -6,6 +6,7 @@ import (
 	"io"
 
 	"github.com/ipld/go-ipld-prime/codec/dagcbor"
+	"github.com/ipld/go-ipld-prime/codec/dagjson"
 	"github.com/ipld/go-ipld-prime/datamodel"
 	"github.com/ipld/go-ipld-prime/node/basicnode"
 	"github.com/ipld/go-ipld-prime/node/bindnode"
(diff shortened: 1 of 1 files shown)


retrievalmarket/bindnodeoptions.go
```diff
@@ -81,6 +82,34 @@ func (sn *CborGenCompatibleNode) MarshalCBOR(w io.Writer) error {
...
+		if tn, ok := node.(schema.TypedNode); ok {
+			node = tn.Representation()
+		}
+	}
+	buf := new(bytes.Buffer)
+	err := dagjson.Encode(node, buf)
+	if err != nil {
+		return nil, err
+	}
+	return buf.Bytes(), nil
```
rvagg, line 99:
```suggestion
	return ipld.Encode(node, dagjson.Encode)
```

retrievalmarket/bindnodeoptions.go
```diff
@@ -81,6 +82,34 @@ func (sn *CborGenCompatibleNode) MarshalCBOR(w io.Writer) error {
...
+}
+
+// UnmarshalJSON is for encoding/json compatibility
+func (sn *CborGenCompatibleNode) UnmarshalJSON(data []byte) error {
+	// convert it to a Node
+	na := basicnode.Prototype.Any.NewBuilder()
+	if err := dagcbor.Decode(na, bytes.NewReader(data)); err != nil {
+		return err
+	}
+	sn.Node = na.Build()
```
rvagg, line 109:
```suggestion
	node, err := ipld.Decode(data, dagjson.Decode)
	if err != nil {
		return err
	}
	sn.Node = node
```

